PR #3, opened 2017-08-23
GitHub auth

Tthomas63:
- [x] Github authentication
- [x] Login, logout, register pages. ( Basic )
- [x] Github pipeline logic for user info
- [x] Custom user model and manager


src/urls.py
```diff
@@ -9,8 +10,11 @@
 
     # Third party
     url('^api/', include('api.urls')),
+    url('', include('social_django.urls', namespace='social')),
```
ckcollab: I think we want to prefix this somehow, like `^social` or something?

src/templates/registration/logged_out.html
```diff
@@ -0,0 +1,15 @@
+{% extends 'base.html' %}
+{% load static %}
+
+<link rel="stylesheet" href="{% static "css/auth.css" %}">
```
ckcollab: we should be including this via some kind of `{% block "extra_headers" %}` that is in the `base.html` `<head>` tag

src/apps/user_auth/models.py
```diff
@@ -0,0 +1,66 @@
+from django.utils import timezone
+
+from django.contrib.auth.base_user import AbstractBaseUser, BaseUserManager
+from django.contrib.auth.models import PermissionsMixin
+from django.db import models
+
+
+class UserManager(BaseUserManager):
```
ckcollab: Let's get rid of this model manager? Should we keep it around instead of just using functions?

src/apps/user_auth/models.py
```diff
@@ -0,0 +1,66 @@
...
+            raise ValueError('Superuser must have is_superuse=True.')
+        return self._create_user(uid, password, **extra_fields)
+
+
+class User(AbstractBaseUser, PermissionsMixin):
+    # Social needs the below setting. Username is not really set to UID.
+    USERNAME_FIELD = 'uid'
+    # Github user attributes.
+    uid = models.CharField(max_length=30, unique=True, blank=True, null=True)
+    login = models.CharField(max_length=50, blank=True, null=True)
```
ckcollab: what is this?

src/apps/user_auth/models.py
```diff
@@ -0,0 +1,66 @@
...
+    avatar_url = models.CharField(max_length=100, null=True, blank=True)
+    url = models.CharField(max_length=100, null=True, blank=True)
+    html_url = models.CharField(max_length=100, null=True, blank=True)
+    name = models.CharField(max_length=100, null=True, blank=True)
+    company = models.CharField(max_length=100, null=True, blank=True)
+    bio = models.CharField(max_length=300, null=True, blank=True)
+    # Any User Attributes
```
ckcollab: can we put newlines on top of these comments?

src/apps/user_auth/pipeline.py
```diff
@@ -0,0 +1,52 @@
...
+        if kwargs:
+            try:
+                response = kwargs.pop("response")
+                for attr in user_attrs:
+                    user_attr_value = getattr(user, attr, None)
+                    github_attr_value = response[attr]
+                    if not user_attr_value or attr not in protected:
+
+                        setattr(user, attr, github_attr_value)
+                setattr(user, 'uid', response['id'])
```
ckcollab: add a comment for this special case?

src/apps/user_auth/pipeline.py
```diff
@@ -0,0 +1,52 @@
...
+                user.save()
+            except AttributeError:
+                print("Attribute not found.")
+        if changed:
+            strategy.storage.user.changed(user)
+
+
+def associate_existing_user(uid, *args, **kwargs):
+    """If there already is an user with the given uid, hand it over to the pipeline"""
```
ckcollab: `an user` -> `a user` I think?

This is very cool

src/settings/base.py
```diff
@@ -100,6 +105,41 @@
...
+    'user_auth.pipeline.associate_existing_user',
+    'user_auth.pipeline.user_details',
+)
+
+SOCIAL_AUTH_GITHUB_SCOPE = ['user']
+SOCIAL_AUTH_STRATEGY = 'social_django.strategy.DjangoStrategy'
+SOCIAL_AUTH_STORAGE = 'social_django.models.DjangoStorage'
+# Change me
+SOCIAL_AUTH_GITHUB_KEY = '1e34985fdb157c961b7c'
+# Change me
```
ckcollab: grab from ENV vars?

src/templates/base.html
```diff
@@ -166,19 +162,20 @@
...
         }
 
         /* Force the footer to stretch all the way to the bottom of viewport */
         .pusher {
-          display: flex;
-          min-height: 100vh;
-          flex-direction: column;
+            display: flex;
+            min-height: 100vh;
+            flex-direction: column;
```
ckcollab: ty for fixing this

src/templates/base.html
```diff
@@ -197,12 +194,16 @@
...
         <a class="item">Chagrade</a>
         <div class="right menu">
-            <div class="item">
-                <a class="ui button">Log in</a>
-            </div>
-            <div class="item">
-                <a class="ui primary button">Sign Up</a>
-            </div>
+            {% if user.is_authenticated %}
+                <div class="item">
```
ckcollab: can we put this conditional inside `<div class="item">` so it's not repeated?

src/templates/registration/base.html
```diff
@@ -0,0 +1,79 @@
...
+<!DOCTYPE html>
+<html>
+<head>
+    <!-- Standard Meta -->
+    <meta charset="utf-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0">
+
+    <!-- Site Properties -->
+    <title>Login Example - Semantic</title>
```
ckcollab: Fix title?

Can we put this file in `templates/base_simple.html` or something, and put `{% extends "base_simple.html" %} in `registration/base.html` 

That way, when we want simple 1 page things with nothing else?

src/templates/registration/base.html
```diff
@@ -0,0 +1,79 @@
...
+    </style>
+    <script>
+        $(document)
+            .ready(function() {
+                $('.ui.form')
+                    .form({
+                        fields: {
+                            email: {
+                                identifier  : 'email',
+                                rules: [
```
ckcollab: can we make this more compact, like put `[{` on the same line and such?

ckcollab:
I tried to auth on the staging server and that didn't work. The callback url was localhost.. hm... wonder how we can do this on staging prs.

src/apps/user_auth/models.py
Tthomas63: `Login` is Github's username, guess we don't really need it now as the pipeline is working. `UID` is from Github as well. We use it to verify that if a user is new or not.

src/apps/user_auth/models.py
Tthomas63: How would we do this with functions vs a model manager?

src/apps/user_auth/pipeline.py
Tthomas63: I believe the rule of thumb for grammar in this case is if the noun begins with a vowel or not. Since U is a vowel it's `an` I believe.

src/apps/user_auth/models.py
```diff
@@ -0,0 +1,69 @@
+from django import contrib
+from django.utils import timezone
+
+from django.contrib.auth.base_user import AbstractBaseUser, BaseUserManager
+from django.contrib.auth.models import PermissionsMixin
+from django.db import models
+
+
+# class UserManager(BaseUserManager):
```
ckcollab: pull this?

src/apps/user_auth/pipeline.py
```diff
@@ -0,0 +1,51 @@
...
+    else:
+        username = strategy.storage.user.get_username(user)
+    return {'username': username}
+
+
+def user_details(user, details, strategy, *args, **kwargs):
+    """Update user details using data from provider."""
+    if user:
+        changed = False  # flag to track changes
+        protected = ('id', 'pk') + tuple(strategy.setting('PROTECTED_USER_FIELDS', []))
```
ckcollab: do we need `protected` if we already have `user_attrs`?

src/apps/user_auth/pipeline.py
ckcollab: https://www.quora.com/Is-it-an-user-or-a-user

Think we can just do "a user" if the vowel sounds like a "y" ??! didn't know that!

Tthomas63:
@ckcollab The checks on this have passed.

ckcollab:
@Tthomas63 :shipit:

PR #25, opened 2017-08-30
trying out drf docs, again

ckcollab:


ckcollab:
Actually the main swagger docs look amazing...! drcdocs doesn't seem maintained, should just use that.

PR #37, opened 2018-01-25
Auth revisit

Tthomas63:
Changes:
 - Template Changes
 - View/Functionallity


src/apps/profiles/forms.py
```diff
@@ -0,0 +1,19 @@
+from django import forms
+from django.contrib.auth.forms import UserCreationForm
+# from django.contrib.auth.models import User
+from .models import User
+
+
+class SignUpForm(UserCreationForm):
```
ckcollab: Pull this, we shouldn't need it I don't think??? Also, just do HTML instead of Form templates, so we don't have to edit placeholders in Python.

src/apps/profiles/views.py
```diff
@@ -1,16 +1,22 @@
...
 from django.http import HttpResponse
 # from django.shortcuts import render
+from django.shortcuts import render, redirect
+
+from .forms import SignUpForm
 
 
 def sign_up(request):
-    return HttpResponse("Not implemented yet!")
+    if request.method == 'POST':
```
ckcollab: There should already be a signup we can link into, I think? but it might be nice to have our own, like to add a "check this box to agree to terms" thing or something..

src/templates/registration/login.html
```diff
@@ -2,66 +2,47 @@
...
-                <div class="field">
-                    <div class="ui left icon input">
-                        <i class="user icon"></i>
-                        <input type="text" name="username" autofocus="" maxlength="254" required="" id="id_username">
+                <div class="ui stacked segment">
+                    <div class="ui field">
+                        <div class="ui left icon input">
+                            <i class="user icon"></i>
+                            <input type="text"
+                                   name="{{ form.username.name }}"
```
ckcollab: shouldn't `name="name"`?

src/templates/registration/signup.html
```diff
@@ -1,44 +1,34 @@
...
+                </h2>
                 {% csrf_token %}
-                {% for field in form %}
-                    <div class="ui stacked segment">
+                <div class="ui stacked segment">
+                    {% for field in form %}
                         <div class="field">
                             <div class="ui left icon input">
-                                {% if field.name == 'username' %}
+                                {% if field.name == 'username' or field.name == 'email' %}
```
ckcollab: yeah like above, just put in the raw HTML here

src/templates/registration/signup.html
```diff
@@ -4,28 +4,69 @@
...
-                                    <i class="write icon"></i>
-                                {% endif %}
-                                {{ field }}
-                            </div>
+                    <div class="ui field">
+                        <div class="ui left icon input">
+                            <i class="user icon"></i>
+                            <input type="text"
+                                   name="{{ form.username.name }}"
+                                   placeholder="{{ form.username.name }}"
```
ckcollab: For placeholder just put the actaul value. For name just put the actual value.

```
name="username" (or whatever)
placeholder="Username"
id="id_username"
```

ckcollab:
Will you also make tests pass??

src/templates/registration/login.html
ckcollab: this still needs a look

src/templates/registration/login.html
```diff
@@ -2,66 +2,47 @@
...
+                                   placeholder="username"
+                                   id="{{ form.username.id_for_label }}"
```
ckcollab: we don't even need this attribute I dont think, pull out as much cruft as you can from here

src/static/stylus/base_template.styl
```diff
@@ -73,6 +73,16 @@
...
     display none
 
+.progress-style
+    margin 0
+    height 0
+    -ms-flex 1 0 auto
+    flex 1 0 auto
+    overflow hidden
+
+.progress24height
```
ckcollab: instead just do it like so at the bottom of `management.tag`, but in regular css, also I don't think we need a new class?

```
.progress
     margin 0
     height 0
     -ms-flex 1 0 auto
     flex 1 0 auto
     overflow hidden
    .bar
        height 24px
```

In regular CSS this owuld be something like:

```
.progress {
    yadda yadda
}
.progress .bar {
    height: 24px;
}
```

Tthomas63: This has been moved

src/templates/registration/login.html
Tthomas63: No? If we simplified it it would be `name="username"` I believe...

src/apps/profiles/views.py
Tthomas63: We decided this is good right? Pretty much the default implementation

src/apps/api/fields.py
```diff
@@ -40,7 +39,7 @@ def to_internal_value(self, named_json_data):
 
             data = ContentFile(decoded_file, name=file_name)
             return data
-        raise ValidationError(_('This is not an base64 string'))
+        raise ValidationError(('This is not an base64 string'))
```
ckcollab: Why pull `_` and why leave in the `()` if you did?

src/apps/api/views/search.py
```diff
@@ -1,11 +1,11 @@
...
-from rest_framework.decorators import api_view
-from rest_framework.response import Response
+# import datetime
+#
+# from django.conf import settings
+# from elasticsearch import Elasticsearch
+# from elasticsearch_dsl import Search
+#
+# from rest_framework.decorators import api_view
+# from rest_framework.response import Response
```
ckcollab: pull please...

src/apps/profiles/views.py
ckcollab: Yeah this seems fine

PR #39, opened 2018-01-26
Competition detail page

Tthomas63:
Changes:
 - Initial Style Setup
 - Pass object via context ( For now )


src/apps/competitions/views.py
```diff
@@ -10,5 +13,16 @@ class CompetitionForm(LoginRequiredMixin, TemplateView):
     template_name = 'competitions/form.html'
 
 
-class CompetitionDetail(LoginRequiredMixin, TemplateView):
+class CompetitionDetail(LoginRequiredMixin, TemplateView, ContextMixin):
     template_name = 'competitions/detail.html'
+
```
ckcollab: You will get competition details via ajax request!

Tthomas63:
Current State:
screen shot 2018-01-28 at 11 03 10 pm
screen shot 2018-01-28 at 11 03 02 pm

ckcollab:
I'm loving it, coming together so nice

ckcollab:
**Current screenshot:** (unless I did something wrong)

image

I would say get rid of the message class everywhere (maybe keep it in a couple spots), let things float out there a bit. Also please make all margins line up (extend to the edge)

-----------

I did a few of these changes and pushed what I did, feel free to undo these changes but i just wanted to push so you could see

PR #61, opened 2018-04-05
52 user management wireframe

### LINKED ISSUE #52: User management wireframe
## Should contain
- A keyword filtering box:
    - Works with emails, username, first name, last name, etc. all relevant fields; use django-filters in DRF for this
- A table of users, sign up date, email, etc. probably limited to the first 50 results
- I don't think pagination will be required

## Questions
* What do we need this for? We haven't had to "ban" anyone from Codalab v1.5, what user management should we need really?

BrendanMyers88:
Resolves #52 

Adds updated Semantic UI CDN call
User Management Wireframe
CSS Call for Modals

### DIFF
--- a/src/static/riot/admin/user_management.tag
+++ b/src/static/riot/admin/user_management.tag
@@ -1,0 +1,109 @@
+<user_management>
+    <!------------------------------------------ HTML ------------------------------------------->
+
+    <div class="row">
+        <h1 class="ui center aligned header">
+            User Management Portal
+        </h1>
+        <div id="search_bar" class="column">
+                <span class="ui search">
+                <span class="ui icon input">
+                    <input class="prompt" type="text" placeholder="Search Users...">
+                    <i class="search icon"></i>
+                </span>
+                <span class="results"></span>
+            </span>
+        </div>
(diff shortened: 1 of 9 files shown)


src/static/riot/admin/user_management.tag
```diff
@@ -0,0 +1,98 @@
...
+        </table>
+    <!-- DELETE USER MODAL -->
+    <div id="delete_user_modal" class="ui mini modal">
+        <div class="header">Delete User JDoe40?</div>
+        <div class="actions">
+            <div class="basic grey ui cancel button">Cancel</div>
+            <div class="ui red approve button">Delete User</div>
+        </div>
+    </div>
+    </div>
```
BrendanMyers88: div indentation?

src/templates/management/user_management.html
```diff
@@ -0,0 +1,7 @@
+{% extends "base.html" %}
+{% load staticfiles %}
+
+{% block content %}
+    <user_management></user_management>
+
```
BrendanMyers88: pull this line

PR #62, opened 2018-04-09
Submission Management Page Wireframe

### LINKED ISSUE #51: Submission management page wireframe
## Should contain

- A user filter field where the input filters on user name, email, etc.
- A Competition typeahead dropdown where you select which competition to filter on
- A table displaying all relevant submissions, pagination probably required
- Table should show details about the submission(s):
    * file name
    * size
    * submitted at
    * started processing at
    * finished processing at
    * Which compute worker did it?
    * What queue was it on
    * What docker image(s) were used?
- The table should have buttons for actions at the top and checkboxes on the left of each submission; there should also be a way to check ALL submissions (even on every subsequent page)
- Available actions:
    * Re-run
    * Stop/cancel
    * Delete

BrendanMyers88:
Resolves #51

### DIFF
--- a/src/static/riot/submissions/submission_management.tag
+++ b/src/static/riot/submissions/submission_management.tag
@@ -1,0 +1,274 @@
+<submission_management>
+    <!-- HTML -->
+    <h1>Submission Management</h1>
+    <div class="ui divider"></div>
+    <div class="ui menu">
+        <div class="item">
+            <div class="ui icon input">
+                <input type="text" placeholder="Search Users...">
+                <i class="user icon"></i>
+            </div>
+        </div>
+        <div class="ui right pushed search item">
+            <div class="ui icon input">
+                <input class="prompt" type="text" placeholder="Search Competitions...">
+                <i class="search icon"></i>
+            </div>
+            <div class="results"></div>
+        </div>
+    </div>
+    <div class="ui">
+        <div class="ui left rail">
+            <div class="ui sticky">
+                <div class="ui vertical labeled icon buttons">
+                    <button class="ui button">
+                        <i class="refresh icon"></i>
+                        Re-run Selected
+                    </button>
+                    <button class="ui button">
+                        <i class="minus circle icon"></i>
+                        Stop Selected
+                    </button>
+                    <button class="ui button">
+                        <i class="trash icon"></i>
+                        Delete Selected
+                    </button>
+                </div>
+                <div class="ui compact hidden info message">
+                    <i class="close icon"></i>
+                    <div class="header">
+                        Success!
+                    </div>
+                    <ul class="list">
+                        <li>Deleted x files</li>
+                        <li>Stopped x files</li>
+                        <li>Re-running x files</li>
+                    </ul>
+                </div>
+        </div>
+    </div>
+    <table class="ui compact celled table">
+        <thead>
+        <tr>
+            <th>
+                <div class="ui checkbox">
+                    <input type="checkbox"><label></label>
+                </div>
+            </th>
+            <th>File Name</th>
+            <th>File Size</th>
+            <th>Submitted at</th>
+            <th>Processing Start</th>
+            <th>Processing End</th>
+            <th>Compute Worker</th>
+            <th>Queue</th>
+            <th>Docker Image</th>
+        </tr>
+        </thead>
+        <tbody>
+        <tr>
+            <td class="collapsing">
+                <div class="ui checkbox">
+                    <input type="checkbox"><label></label>
+                </div>
+            </td>
+            <td></td>
+            <td></td>
+            <td></td>
+            <td></td>
+            <td></td>
+            <td></td>
+            <td></td>
+            <td></td>
+        </tr>
+        <tr>
+            <td class="collapsing">
+                <div class="ui checkbox">
+                    <input type="checkbox"><label></label>
+                </div>
+            </td>
+            <td></td>
+            <td></td>
+            <td></td>
+            <td></td>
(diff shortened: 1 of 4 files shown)


src/static/riot/submissions/submission_management.tag
```diff
@@ -0,0 +1,266 @@
+<submission_management>
+    <!-- HTML -->
```
ckcollab: don't really need this comment

ckcollab:
image

When you scroll, the top part scrolls as well. Can we put this in some kind of container so only that container scrolls, not the top part or pagination, maybe? Might not be do-able easily, and this is certainly fine for now.

src/static/riot/submissions/submission_management.tag
```diff
@@ -0,0 +1,266 @@
...
+        </div>
+        <div class="ui right pushed search item">
+            <div class="ui icon input">
+                <input class="prompt" type="text" placeholder="Search Competitions...">
+                <i class="search icon"></i>
+            </div>
+            <div class="results"></div>
+        </div>
+    </div>
+    <div id="horizButtons" class="ui horizontal icon buttons">
```
ckcollab: I think we should get in the habit of naming all IDs [snake_case](https://en.wikipedia.org/wiki/Snake_case) just to keep a consistent style

src/static/riot/submissions/submission_management.tag
```diff
@@ -0,0 +1,266 @@
...
+                        <a class="item">3</a>
+                        <a class="item">4</a>
+                        <a class="icon item"><i class="right chevron icon"></i></a>
+                    </div>
+                </th>
+            </tr>
+            </tfoot>
+        </table>
+    </div>
+    <!-- Javascript -->
```
ckcollab: Hmm, I like the idea of these, I think we can get the same idea with just space? I think 2 spaces between HTML - JS - CSS should be fine. What do you think?

Also, commented "sections" in JS we'd normally do something like...

```js
        /* ----------------------------------------------------------------------------
         Initializer
         ----------------------------------------------------------------------------*/

... 
        /* ----------------------------------------------------------------------------
         Event handlers
         ----------------------------------------------------------------------------*/

```

src/static/riot/submissions/submission_management.tag
```diff
@@ -0,0 +1,266 @@
...
+    <script>
+        $(document).ready(function () {
+            $('.ui.sticky')
```
ckcollab: should we name this more specifically in case we have other `.ui.sticky` things?

src/static/riot/submissions/submission_management.tag
```diff
@@ -0,0 +1,266 @@
...
+                .sticky({
+                    context: '.ui.compact.celled.table',
+                    offset: 75
+                })
+        })
+    </script>
+    <!-- CSS -->
```
ckcollab: same thing here with comment, prob good to just do 2 spaces

src/templates/base.html
```diff
@@ -130,8 +130,10 @@
         {% block top_div_extra_content %}{% endblock %}
     </div>
     <div class="ui container grid">
-        {% block content %}
-        {% endblock %}
+        <div class="row">
```
ckcollab: Ah, great catch, we do need row here so it doesn't futz up margins?

ckcollab:
Also, did you try "Stickying" the buttons to the left side of the table?

src/static/riot/submissions/submission_management.tag
BrendanMyers88: Removed comment

src/static/riot/submissions/submission_management.tag
BrendanMyers88: changed id case to snake_case

src/static/riot/submissions/submission_management.tag
BrendanMyers88: removed comment

src/static/riot/submissions/submission_management.tag
BrendanMyers88: changed to class to 'ui sticky leftrail'

src/static/riot/submissions/submission_management.tag
BrendanMyers88: removed comment

src/templates/base.html
BrendanMyers88: Correct, it keeps margins working properly

ckcollab:
2 things:

## Breakpoint width
I think maybe we should change the "breakpoint" for the buttons to be like ~1150 instead of 1200?

image

Very small detail, but hides stuff a bit.

## Tooltips

I think they should pop out to the right now, instead of top/bottom?

src/static/riot/submissions/submission_management.tag
```diff
@@ -0,0 +1,279 @@
...
+                </th>
+            </tr>
+            </tfoot>
+        </table>
+    </div>
+
+
+    <script>
+        $(document).ready(function () {
+            $('.ui.sticky.leftrail')
```
ckcollab: why don't we remove this class and instead do...

`$('#leftrail .ui.sticky')`?

ckcollab:
:shipit: this is ready to go, just leaving it up so Isabelle can see it

ckcollab:
Last thing I suppose is resolving this:

> `/static/riot/submissions/submit.tag` was old Submission Management tag file. Please check and see if we want to keep the file or remove it from the project.

Where did we land on that? We want to pull it, I think?

PR #66, opened 2018-04-30
Makes the height of the jumbotron relative to the viewport size

### LINKED ISSUE #65: Homepage/Jumbotron height are not relative to the viewport
Because the homepage height is not relative to the viewport, the jumbotron does not display to the bottom of the page. This can leave a gap depending on the size of the screen viewing the Codalab homepage.

BrendanMyers88:
Removes the whitespace from the bottom of the homepage and fills the screen with the jumbotron.

Resolves #65

### DIFF
--- a/src/static/stylus/home.styl
+++ b/src/static/stylus/home.styl
@@ -1,2 +1,8 @@
 #jumbo_logo
-    margin-left -45px
+  margin-left -45px
+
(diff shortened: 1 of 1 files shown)


BrendanMyers88:
Decided to go a different direction on styling.

PR #71, opened 2018-08-20
Landing page redesign

BrendanMyers88:
Codalab Landing Page Re-Design:

- Changed color scheme
- Buttons are nicely styled
- Menus are lined up and mobile friendly
- Changed Font of Codalab Logo to Futura LT Light
- Removed Fixed menu due to bugginess
- Sidebar is working
- Dropdown is working


src/static/riot/competitions/tile/competition_tile.tag
```diff
@@ -63,15 +63,20 @@
             margin-bottom 35px !important
 
         .main-wrapper
-            border-left 5px solid lightgreen
+            border-left 5px solid #c7402d
```
ckcollab: hmm, should we define these colors somewhere..............

src/static/stylus/base_template.styl
```diff
@@ -5,20 +5,54 @@
  */
 #bg
     z-index -1
-    background rgba(22, 12, 160, 0.9)
     position absolute
     top 0
     bottom 0
     left 0
     right 0
+    background rgba(19, 48, 70, 0.9)
```
ckcollab: same with this color, if we use it often

src/static/stylus/base_template.styl
```diff
@@ -5,20 +5,54 @@
...
+
+#style_button:hover
+    color #3f3f3f!important
+    background-color rgba(255, 255, 255, .65)!important
+    box-shadow none!important
+    .icon
+        opacity 1 !important
+
+.ui.dropdown
+    position absolute!important
```
ckcollab: can we get away from the `!important`'s? I'm not sure we can.. may have to research how selector precedence works.

Also, should have a space between the style and `!important`

src/templates/pages/home.html
```diff
@@ -31,7 +33,22 @@ <h2>Accelerating reproducible computational research.</h2>
     {#                </div>#}
     {#            </div>#}
     {#        </div>#}
-
+    <script>
+    // Prevent buttons from stacking in mobile view
+    $(document).ready(function () {
+        $(window).resize(function resize() {
```
ckcollab:
```javascript
function resize() {
```

Can just be

```javascript
function {
```

Makes sense that it's an "anonymous" function since we're just passing it in here and not reusing it anywhere.

src/templates/pages/home.html
```diff
@@ -31,7 +33,22 @@ <h2>Accelerating reproducible computational research.</h2>
...
+            if ($(window).width() < 390) {
+                return $('.pages')
+                    .removeClass('huge')
+                    .addClass('large');
+            }
+
+            $('.pages')
+                .removeClass('large')
+                .addClass('huge');
+        }).trigger('resize');
```
ckcollab: Ahh! That's neat, instead of having to make resize a function and re-using/calling it, `trigger` is cool, not sure I've seen that before..

src/static/stylus/base_template.styl
BrendanMyers88: Added Color Reference to base_template.styl

src/static/riot/competitions/tile/competition_tile.tag
BrendanMyers88: Added Color Reference to base_template.styl

PR #108, opened 2019-01-13
Task comp upload problems (for quick review)

ckcollab:


src/apps/competitions/tasks.py
```diff
@@ -182,39 +184,87 @@ def unpack_competition(competition_dataset_pk):
...
+                                    raise CompetitionUnpackingException(f'Cannot find dataset: "{file_name}" for task: "{new_task["name"]}"')
+                            new_phase["tasks"].append(new_task)
+                        elif type(task) is str:
+                            # lookup as UUID
+                            print("I am a string")
+                            new_phase["tasks"].append(task)
+                        else:
+                            print(f"\nERROR invalid task structure: \n\n type: {type(task)},\n task: {task}")
+
+                else:
```
ckcollab: If we don't have "Tasks" then we should treat it as a normal competition?

ckcollab: Ah maybe that's all that's happening, I was confused by these comments (I think I added them..!)

```python

                            # verify as UUID?

                            # TODO UUIDs can't have '.'s acceptable form of validating file vs uuid?
```

src/apps/competitions/tasks.py
```diff
@@ -182,39 +184,87 @@ def unpack_competition(competition_dataset_pk):
...
+                        if type(task) is dict:
+                            print('This is a dict')
+                            # handle as dictionary like normal
+                            new_task = {
+                                'name': task['name'],
+                                'description': task['description'] if 'description' in task else None,
+                                'created_by': creator.id,
+                            }
+                            # TODO this is where we will process modules eventually. For now just data sources
+                            file_types = ['reference_data', 'scoring_program']
```
ckcollab: we have to be careful about re-using this `file_types` name, might mess up the other case/something?

Issue #91, opened 2018-11-27
Factories for testing


ckcollab:
Some way to generate:
- [x] 10 new users
- [ ] 10 new submissions
    * creating users if not provided
    * competition must be provided
    * optionally automatically execute submissions
    * optionally automatically attempt to add to leaderboard
- [ ] 10 new competitions 
    * must create phase/etc. required fields
    * optionally creating users
    * optionally creating submissions

Use [factory boy](https://github.com/FactoryBoy/factory_boy)

jimmykodes:
- [ ] Create `generate_data` management command implementing these factories.

PR #116, opened 2019-01-28
Task solution and v1.5 to 2 features (for review)

jimmykodes:
Oh God, so many changes. Probably should have built v1.5 -> v2 stuff not on top of tasks and solutions for a clearer PR. It's gross looking, but at least a cursory review point? PR into branch I based my stuff on for clarity's sake.


docker-compose.yml
```diff
@@ -77,7 +77,7 @@ services:
   #   Database Service
   #-----------------------------------------------
   db:
-    image: postgres:10.6-alpine
+    image: postgres:11.1-alpine
```
ckcollab: was this necessary? will it break databases from 10.6?

pytest.ini
```diff
@@ -1,3 +1,3 @@
 [pytest]
-DJANGO_SETTINGS_MODULE = settings.test
+addopts = --ds=settings.test
```
ckcollab: Maybe add a comment here?

```
# DJANGO_SETTINGS_MODULE = settings.test was not working, using this to manually set
# django settings module
addopts = --ds=settings.test
```

src/apps/api/urls.py
```diff
@@ -16,6 +16,8 @@
 router.register('datasets', datasets.DataViewSet)
 router.register('data_groups', datasets.DataGroupViewSet)
 router.register('leaderboards', leaderboards.LeaderboardViewSet)
+router.register('tasks', tasks.TaskViewSet)
+router.register('tasksearch', tasks.TaskViewSetSimple)
```
ckcollab: What's the difference between `TaskViewSet` vs `TaskViewSetSimple`?

src/apps/api/urls.py
```diff
@@ -43,7 +45,7 @@
     url(r'swagger(?P<format>\.json|\.yaml)$', schema_view.without_ui(cache_timeout=0), name='schema-json'),
     path('swagger/', schema_view.with_ui('swagger', cache_timeout=0), name='schema-swagger-ui'),
     path('redoc/', schema_view.with_ui('redoc', cache_timeout=0), name='schema-redoc'),
-
+    path('can_make_submission/<phase_id>/', submissions.can_make_submission),
```
ckcollab: So thinking about this more, can we easily do this step on getting the `SAS` url? Hmmm....

src/apps/api/views/submissions.py
```diff
@@ -89,3 +90,17 @@ def upload_submission_scores(request, submission_pk):
         )
 
     return Response()
+
+
+@api_view(('GET',))
+def can_make_submission(request, phase_id):
+    # TODO: Check that user is in competition
```
ckcollab: todo to hit, but still TBD on exactly where to do this (re: SAS discussion above)

src/apps/api/views/tasks.py
```diff
@@ -0,0 +1,58 @@
...
+
+from api.serializers import tasks as serializers
+from tasks.models import Task
+
+
+class TaskViewSet(ModelViewSet):
+    queryset = Task.objects.all()
+    serializer_class = serializers.TaskSerializer
+    filter_backends = (DjangoFilterBackend, SearchFilter)
+    # filter_fields = ('type', 'name', 'key',)
```
ckcollab: pull cruft?

src/apps/api/views/tasks.py
```diff
@@ -0,0 +1,58 @@
...
+    filter_backends = (DjangoFilterBackend, SearchFilter)
+    search_fields = ('name',)
+
+    def get_queryset(self):
+        filters = Q(is_public=True) | Q(created_by=self.request.user)
+        qs = Task.objects.filter(filters)
+        return qs
+
+    def get_serializer_context(self):
+        # Have to do this because of docs sending blank requests (?)
```
ckcollab: "docs" sending blank requests?

src/apps/api/views/tasks.py
```diff
@@ -0,0 +1,58 @@
...
+    # filter_fields = ('type', 'name', 'key',)
+    search_fields = ('name', 'description',)
+
+    def get_queryset(self):
+        filters = Q(is_public=True) | Q(created_by=self.request.user)
+        qs = Task.objects.filter(filters)
+        return qs
+
+    def get_serializer_context(self):
+        # Have to do this because of docs sending blank requests (?)
```
ckcollab: Docs sending blank requests?

src/apps/api/views/tasks.py
```diff
@@ -0,0 +1,58 @@
...
+    queryset = Task.objects.all()
+    serializer_class = serializers.TaskSerializer
+    filter_backends = (DjangoFilterBackend, SearchFilter)
+    # filter_fields = ('type', 'name', 'key',)
+    search_fields = ('name', 'description',)
+
+    def get_queryset(self):
+        filters = Q(is_public=True) | Q(created_by=self.request.user)
+        qs = Task.objects.filter(filters)
+        return qs
```
ckcollab: Probably should be:

```python
return Task.objects.filter(Q(is_public=True) | Q(created_by=self.request.user))
```

src/apps/competitions/tasks.py
```diff
@@ -94,6 +96,30 @@ class CompetitionUnpackingException(Exception):
...
+            created_by=creator,
+            type=file_type,
+            name=f"{file_type} @ {now().strftime('%m-%d-%Y %H:%M')}",
+            was_created_by_competition=True,
+        )
+        new_dataset.data_file.save(os.path.basename(file_path), File(open(file_path, 'rb')))
+        return new_dataset.key
+    elif len(file_name) in (32, 36):
+        # UUID are 32 or 36 characters long
+        # TODO send error message if invalid UUID or invalid filename
```
ckcollab: should we do more here? Can make `django-unique-uuidfield` library !?

src/apps/competitions/views.py
```diff
@@ -20,6 +21,18 @@ class CompetitionDetail(LoginRequiredMixin, DetailView):
     queryset = Competition.objects.all()
     template_name = 'competitions/detail.html'
 
+    def get_object(self, *args, **kwargs):
+        competition = super().get_object(*args, **kwargs)
+
+        if self.request.user.is_superuser or self.request.user == competition.created_by or self.request.user in competition.collaborators.all():
```
ckcollab: Hmm, this feels weird, what do you think of this style?

```python
is_creator = self.request.user.is_superuser or self.request.user == competition.created_by
is_collaborator = self.request.user in competition.collaborators.all()
valid_secret_key = self.request.GET.get('secret_key') == str(competition.secret_key)
if is_creator or is_collaborator or competition.published or valid_secret_key:
    return competition
raise Http404()
```

src/apps/api/urls.py
jimmykodes: I know I created `Simple` for working with the Phase Modal, trying to get Select2 to work. Since we kinda jumped ship on that one and put me to work on v1.5 -> v2 stuff (i.e. things I understood how to do), I don't think this ever got optimized (or maybe even working), so I'm not super sure `Simple` is necessary at all.

src/apps/api/views/submissions.py
jimmykodes: This todo also require implementing `CompetitionParticipant` things I think.

src/apps/api/urls.py
ckcollab: cool, gotchya, could you confirm we don't need it any more and pull it or at least comment it out?

src/apps/api/views/submissions.py
ckcollab: Ahh, okay, made #118 to track that

src/apps/api/views/tasks.py
jimmykodes: https://github.com/codalab/competitions-v2/blob/d72c9dc418734e95ac7a7a92d985e8cb61291cba/src/apps/api/views/competitions.py#L51

Pulled it from here, didn't really know why it was there, but didn't touch it so as not to break things?

src/apps/api/urls.py
jimmykodes: So this was created for Select2 to dynamically populate its dropdown via searching. It needs a serializer that has specific fields. Could _maybe_ be done with a get serializer context? but I don't know how we would differentiate which serializer gets called.

src/apps/api/views/tasks.py
jimmykodes: - [x] done

src/apps/api/views/tasks.py
jimmykodes: - [x] done

src/apps/competitions/tasks.py
jimmykodes: I don't think this is a matter of uniqueness, this TODO is more about a filename that is 32 characters long, but the file isn't present. It treats it like a UUID but then doesn't raise any error if it can't find that UUID in the database anywhere. However, I would also love to write a `django-unique-uuidfield` package, since that is not something I have ever attempted before.

src/apps/competitions/views.py
jimmykodes: - [x] done

pytest.ini
jimmykodes: - [x] done

src/apps/competitions/tasks.py
ckcollab: Ah, kk, can track that here #119

docker-compose.yml
jimmykodes:
```
2019-01-28 23:33:08.684 UTC [1] FATAL:  database files are incompatible with server 
2019-01-28 23:33:08.684 UTC [1] DETAIL:  The data directory was initialized by PostgreSQL version 11, which is not compatible with this version 10.6.
```
Error I get if I try and use 10.6. Not totally sure why? something to do with my local postgres?

src/apps/api/urls.py
ckcollab: Hmm, do you think you could make a custom view with `@action` or switch the serializer in `get_serializer_class`? Maybe this is an OK way for now.. hmm...

jimmykodes: Possibly, I'm not super sure what the logic of it all would be, especially if we want to search things different on a different screen? like task management screen. I can tag the serializers and views with a TODO so we think about it when we jump back in to this portion of things?

ckcollab: 👍

src/apps/api/views/tasks.py
ckcollab: Gross. mind adding `TODO: why are we doing this? is this still a problem?` or some-such?

docker-compose.yml
jimmykodes: You mean uninstall postgres and re-install an older version? doesn't that kind of defeat the purpose of docker? plus that will break anything else I'm using postgres for locally? Or am i misunderstanding you?

src/apps/api/views/tasks.py
jimmykodes: 👍🏻

src/apps/competitions/tests/test_submissions.py
```diff
@@ -0,0 +1,121 @@
...
+            created_by=self.user,
+        )
+        self.phase = Phase.objects.create(
+            competition=self.competition,
+            index=0,
+            start=now(),
+            name='Test Phase',
+        )
+
+    def make_submission(self, user=None, phase=None, data=None, submission_status=None, submission_date=None, submission_name=None):
```
ckcollab: NOTE: Should move this somewhere common and/or use factory boy, when you implement this issue #91

src/apps/competitions/tests/test_submissions.py
```diff
@@ -0,0 +1,121 @@
...
+            assert False, "This counted failed submissions"
+
+    def test_max_per_day_not_counting_previous_days_submissions(self):
+        self.set_max_submissions(per_day=1)
+        yesterday = now() - timedelta(days=1)
+        self.make_submission(submission_date=yesterday)
+        try:
+            self.make_submission()
+        except PermissionError:
+            assert False, "This counted yesterday's submissions"
```
ckcollab: this is pretty!

docker-compose.yml
ckcollab: I think there's a miscommunication; we can drop this version down, merge this, and fix your machine tomorrow

ckcollab:
:shipit:

PR #125, opened 2019-02-01
added factory boy factories and generate_data command

jimmykodes:


src/apps/commands/management/commands/generate_data.py
```diff
@@ -0,0 +1,59 @@
...
+                        SubmissionFactory(phase=phase, owner=random.choice(users))
+                        submissions += 1
+
+        for user in users:
+            competitions = []
+            for submission in user.submission.all().select_related('phase__competition'):
+                competitions.append(submission.phase.competition)
+            competitions = set(competitions)
+            for competition in competitions:
+                CompetitionParticipant.objects.create(
```
ckcollab: Would be nice to use a factory here? Can use the same factory in tests

src/apps/commands/management/commands/generate_data.py
```diff
@@ -0,0 +1,59 @@
...
+        print(f'Created {len(users)} users.')
+        print(f'Created {competition_count} competitions.')
+        print(f'Created {phases} phases.')
+        print(f'Created {submissions} submissions.')
+
+    @staticmethod
+    def create_competition_with_phases(user, nb_phases):
+        competition = CompetitionFactory(created_by=user)
+        for i in range(nb_phases):
+            PhaseFactory(competition=competition, index=i)
```
ckcollab: Rename `nb_phases` to `phase_count` or something, maybe?

src/factories.py
```diff
@@ -0,0 +1,70 @@
...
+        model = User
+        django_get_or_create = ('username',)
+
+    username = factory.Faker('user_name')
+    email = factory.LazyAttribute(lambda o: f"{o.username}@example.com")
+    password = factory.PostGenerationMethodCall('set_password', 'admin')
+
+    @post_generation
+    def super_user(self, created, extracted, **kwargs):
+        if extracted:
```
ckcollab: what does `extracted` do here? neat

src/factories.py
```diff
@@ -0,0 +1,70 @@
...
+            self.is_staff = True
+
+
+class CompetitionFactory(DjangoModelFactory):
+    class Meta:
+        model = Competition
+
+    title = factory.Sequence(lambda n: f'Competition {n}')
+    created_by = factory.SubFactory(UserFactory)
+    published = factory.LazyAttribute(lambda n: random.choice([True, False]))
```
ckcollab: very cool!

src/apps/commands/management/commands/generate_data.py
```diff
@@ -0,0 +1,59 @@
...
+            '-s'
+            '--size',
+            type=int,
+            help='Determines the size of data to be created, i.e. 5 users, each with 5 comps with 5 phases with 5 submissions.'
+                 'This defaults to 5')
+
+    def handle(self, *args, **kwargs):
+        size = kwargs['size'] if 'size' in kwargs else 5
+        users = [UserFactory(username='admin', super_user=True) if i == 0 else UserFactory() for i in range(size)]
+        competition_count = 0
```
ckcollab: do we need all of these counters? aren't they deterministic based off of input size?

src/apps/commands/management/commands/generate_data.py
ckcollab: Couldn't you make a `CompetitionParticipant` factory that also makes a competition, phase, etc. if not given? Hmm... may need to think about this more

src/factories.py
jimmykodes: extracted is whatever you supply it in the instantiation (if that's what its called) of a factory:
`UserFactory(super_user=True)`
`True` is the value of extracted
Can also do stuff like:
`UserFactory(super_user__is_staff=False)` or something like that and then `is_staff=False` would be present in `kwargs`.

src/apps/commands/management/commands/generate_data.py
jimmykodes: Yea, I can pull them. I had them there to verify the number of things being generated was as I expected. Since it is nested, it is exponential (i.e. size 5 is 5 users but 625 submissions), and things weren't coming out right, made for helpful error checking. But also I liked seeing how many of each thing were actually created, mostly for funziez.

src/factories.py
jimmykodes: Could change `extracted` to `value` or something to make it clearer?

src/apps/commands/management/commands/generate_data.py
jimmykodes: - [x] done, this ends up being a lot cleaner just using a factory for this.

src/apps/commands/management/commands/generate_data.py
jimmykodes: - [x] done

src/apps/commands/management/commands/generate_data.py
jimmykodes: - [x] done

PR #149, opened 2019-03-02
Dog fooding dummy competition

ckcollab:


src/static/riot/tasks/management.tag
```diff
@@ -64,28 +106,63 @@
...
+                    toastr.error("Could not load tasks")
+                })
+        }
+
+
+        self.search_my_tasks = function () {
+            var filter = self.refs.search_mine.value
+
+            if (filter !== "" && filter.length < 3) {
+
```
ckcollab: can just return here and then we don't need the `else` code indented? also, could maybe reduce the delay code a bit...

```delay(() => self.update_my_tasks({search: filter}), 100)```

File: src/static/riot/competitions/form/_phases.tag [hunk 2]
```diff
@@ -46,8 +46,8 @@
         <div class="content">
             <div class="ui top pointing secondary menu">
                 <a class="active item" data-tab="phase_details">Phase details</a>
-                <a class="item" data-tab="phase_datasets">Datasets</a>
-                <a class="item" data-tab="phase_task">Tasks</a>
+                <a class="item" data-tab="phase_datasets" if="{ current_phase_style == 'dataset' }">Datasets</a>
+                <a class="item" data-tab="phase_task" if="{ current_phase_style == 'task' }">Tasks</a>
```
jimmykodes: I think these need to be show instead of if? or something else needs to get tweaked so the tabs actually work. Can probably handle as a part of the competition edit task

ckcollab: Unless it broke in the mean time, was working -- need to clear cache maybe?

src/static/riot/competitions/management.tag
```diff
@@ -1,13 +1,13 @@
...
+            <h1 style="float: left; display: inline-block;">Competition Management</h1>
             <a class="ui right floated green button" href="{ URLS.COMPETITION_UPLOAD }">
                 <i class="upload icon"></i> Upload
             </a>
-            <a class="ui right floated green button" href="{ URLS.COMPETITION_ADD }">
+            <!--<a class="ui right floated green button" href="{ URLS.COMPETITION_ADD }">
                 <i class="add square icon"></i> Create
-            </a>
+            </a>-->
         </div>
```
jimmykodes: Should we uncomment this before we pull it in to develop?

src/static/riot/competitions/form/_phases.tag
jimmykodes: The menu item changes, I just can't actually change to the tab.

ckcollab: got it, you're right, `show` works

src/static/riot/competitions/management.tag
ckcollab: good call, done

jimmykodes:
Is this ready to merge now?

ckcollab:
I believe so, :shipit: should be fine

PR #159, opened 2019-03-09
ChaRepo Mocks (Table styling, etc)

Tthomas63:
- Mocks New ChaRepo Style Dataset management table
  * Highlighting for auto-created
  * Checkbox to hide/show auto-created
  * Pagination
  * Form as modal

screencapture-localhost-datasets-2019-03-08-17_51_16


src/apps/api/pagination.py
```diff
@@ -0,0 +1,7 @@
+from rest_framework.pagination import PageNumberPagination
+
+
+class DatasetPagination(PageNumberPagination):
+    page_size = 10
```
ckcollab: can we put these straight on the view instead of having to do this subclass thing? If we have to do it this way, can we rename this `BasicPagination` so we can make it more clear it's for re-use?

src/static/js/ours/client.js
```diff
@@ -125,8 +125,8 @@ CODALAB.api = {
     /*---------------------------------------------------------------------
          Datasets
     ---------------------------------------------------------------------*/
-    get_datasets: function (query, type) {
-        return CODALAB.api.request('GET', URLS.API + `datasets/?q=${query || ''}&type=${type || ''}`)
+    get_datasets: function (query, type, page) {
+        return CODALAB.api.request('GET', URLS.API + `datasets/?q=${query || ''}&page=${page || 1}&type=${type || ''}`)
```
ckcollab: we should be using `$.param()` or what-have-you, like you can see here: https://github.com/codalab/chahub/blob/develop/src/static/js/ours/client.js#L17

src/static/riot/datasets/management.tag
```diff
@@ -1,9 +1,82 @@
...
+                </tr>
+               <!-- -->
+                </tfoot>
+            </table>
+        </div>
+    </div>
+
 
-    <div class="ui two column grid">
+    <div id="dataset_modal" class="ui modal">
```
ckcollab: Should we just use a `ref` instead of `id`? That way we don't have to do the jquery lookup each time, more riotjs-ee, etc.

src/apps/chahub/tests/test_chahub_mixin.py
```diff
@@ -0,0 +1,103 @@
+import datetime
+import mock
```
jimmykodes: Should this be `from unittest import mock`?

src/apps/chahub/tests/test_chahub_mixin.py
```diff
@@ -0,0 +1,103 @@
...
+from django.conf import settings
+from django.http.response import HttpResponseBase
+from django.test import TestCase
+# from django.test.utils import override_settings
+
+from apps.authenz.models import ClUser
+from apps.web.models import CompetitionSubmission, Competition, CompetitionPhase, CompetitionParticipant, \
```
jimmykodes: Where are these apps actually at? Tests can't seem to find them?

src/apps/chahub/tests/test_chahub_utils.py
```diff
@@ -0,0 +1,44 @@
+import datetime
+import mock
```
jimmykodes: import statement fix?

src/apps/chahub/tests/test_chahub_utils.py
```diff
@@ -0,0 +1,44 @@
...
+from django.conf import settings
+
+from django.test import TestCase
+from django.test.utils import override_settings
+
+from apps.authenz.models import ClUser
+from apps.web.models import CompetitionSubmission, Competition, CompetitionPhase, CompetitionParticipant, \
+    ParticipantStatus
+from apps.web.tasks import send_chahub_general_stats
```
jimmykodes: Same question as before, where are these apps?

src/apps/commands/management/commands/generate_data.py
```diff
@@ -10,15 +10,25 @@ class Command(BaseCommand):
     def add_arguments(self, parser):
         # Optional argument
         parser.add_argument(
-            '-s'
+            '-n',
+            '--no-admin',
+            type=bool,
+            help='Whether to create an admin'
```
jimmykodes: Why does this default to true? shouldn't it default to false, and then we add the `--no-admin` flag if we _don't_ want an admin account? If we want the default behavior to be not creating an admin account, we should rename this command to `--create-admin` or something so it is additive instead of exclusive, and so it is much more intuitive.

src/apps/datasets/models.py
```diff
@@ -60,12 +63,70 @@ def save(self, *args, **kwargs):
...
+        from competitions.models import Phase
+        phases = Phase.objects.filter(Q(ingestion_program=self) | Q(input_data=self) | Q(reference_data=self) | Q(
+            scoring_program=self)).prefetch_related('competition')
+        print(tasks)
+        print(phases_from_tasks)
+        print(phases)
+        task_competitions = [phase.competition.pk for phase in phases_from_tasks if phase.competition]
+        phase_competitions = [phase.competition.pk for phase in phases if phase.competition]
+        competition_set = list(set(task_competitions + phase_competitions))
+        is_used = bool(competition_set)
```
jimmykodes: Ditch the print statements.
Also, can simplify this logic a lot to something like this:
```
competitions = Competition.objects.filter(
    Q(phases__ingestion_program=self) |
    Q(phases__input_data=self) | 
    Q(phases__reference_data=self) | 
    Q(phases__scoring_program=self) |
    Q(phases__tasks__ingestion_program=self) |
    Q(phases__tasks__input_data=self) | 
    Q(phases__tasks__reference_data=self) | 
    Q(phases__tasks__scoring_program=self) |
).distinct()
```

src/apps/datasets/models.py
```diff
@@ -60,12 +63,70 @@ def save(self, *args, **kwargs):
...
         return self.name or ''
 
+    def get_chahub_endpoint(self):
+        return "datasets/"
+
+    def clean_chahub_data(self, temp_data):
+        data = temp_data
+        for key in list(data):
+            if not data[key] or data[key] == '':
+                data.pop(key, None)
```
jimmykodes: If you add a `continue` here, you don't need to do the `data.get()`s down below? Since you don't actually need to check the other `if`s if you have popped it from the data.

src/apps/datasets/models.py
```diff
@@ -60,12 +63,70 @@ def save(self, *args, **kwargs):
...
+            if isinstance(data.get(key), datetime.datetime):
+                data[key] = data[key].isoformat()
+            if isinstance(data.get(key), uuid.UUID):
+                data[key] = str(data[key])
+        return data
+
+    def get_chahub_data(self):
+        print("SENDING A DATASETS WHOOOOOOOOO WWEEEEEEEEEEE")
```
jimmykodes: pull

src/apps/profiles/models.py
```diff
@@ -38,3 +52,43 @@ def get_full_name(self):
...
+        if associations.count() > 0:
+            return associations.first().uid
+        return None
+
+    def get_chahub_endpoint(self):
+        return "profiles/"
+
+    def clean_chahub_data(self, temp_data):
+        data = temp_data
+        for key in list(data):
```
jimmykodes: Would it make more sense to only _add_ what we want to data, rather than add everything to data and try and clean it? i.e.
```python
data = {}
for key, item in temp_data.items():
    if not item or key in PROFILE_DATA_BLACKLIST:
        continue
    elif key == 'details':
        item = self.clean_chahub_data(item)
    elif isinstance(item, datetime.datetime):
        item = item.isoformat()
    data[key] = item
```

src/static/js/ours/client.js
jimmykodes: why not
```javascript
    get_datasets: function (filters) {
        return CODALAB.api.request('GET', `${URLS.API}datasets/`, filters)
    },
```
?

src/static/riot/datasets/management.tag
```diff
@@ -199,6 +246,20 @@
                 // Filters
                 var search = self.refs.search.value.toLowerCase()
                 var type = self.refs.type_filter.value
+                var auto_created = self.refs.auto_created.checked
+
+                console.log("@@@@@@@@@")
+                console.log(auto_created)
```
jimmykodes: pull

src/static/riot/datasets/management.tag
```diff
@@ -334,14 +417,26 @@
             var metadata = get_form_data(self.refs.form)
             delete metadata.data_file  // dont send this with metadata
 
+
+            console.log("@@@@@@@@")
+            console.log(metadata)
```
jimmykodes: pull

src/apps/chahub/utils.py
```diff
@@ -0,0 +1,48 @@
...
+                'X-CHAHUB-API-KEY': settings.CHAHUB_API_KEY,
+            }
+        }
+        if update:
+            # return requests.patch(data=data, **kwargs)
+            return requests.put(data=data, **kwargs)
+        else:
+            return requests.post(data=data, **kwargs)
+    except requests.ConnectionError:
+        return None
```
jimmykodes: We should probably return a more accurate error message than simply `None`. And we are trying to catch this error twice (that I have seen so far), so in the second place we won't actually get the error message.
see: src/apps/chahub/tasks.py  Line 22.
Let's add this as a `# TODO:` here. I'll plan on addressing it when I start re-writing tests.

src/apps/datasets/models.py
```diff
@@ -60,12 +63,64 @@ def save(self, *args, **kwargs):
...
+    @property
+    def in_use(self):
+        from tasks.models import Task
+        tasks = Task.objects.filter(
+            Q(ingestion_program=self) |
+            Q(input_data=self) |
+            Q(reference_data=self) |
+            Q(scoring_program=self)
+        )
+        phases_from_tasks = [phase for task in tasks for phase in task.phases.all()]
```
jimmykodes: why are we querying tasks to filter back up to competitions when we can query competitions based on tasks?

src/apps/chahub/utils.py
Tthomas63: Added two do's on both ends for now. We need to come up with a way to standardize that so we all know which errors should bubble out, and which should not.

PR #194, opened 2019-05-20
fixed phase ordering and pages on competition detail

jimmykodes:


src/static/riot/competitions/detail/_tabs.tag
```diff
@@ -246,47 +173,23 @@
...
-                CODALAB.api.toggle_competition_publish(competition.id)
+            self.toggle_competition_publish = function () {
+                let comp_id = self.competition.id
+                CODALAB.api.toggle_competition_publish(comp_id)
                     .done(function (data) {
                         var published_state = data.published ? "published" : "unpublished"
                         toastr.success(`Competition has been ${published_state} successfully`)
                         self.update()
-                        CODALAB.api.get_competition(self.competition.id)
+                        CODALAB.api.get_competition(comp_id)
```
ckcollab: hmm, i think the `comp_id` temp holder is kinda gross here? would prefer `self.competition.id`?

src/static/riot/competitions/detail/_tabs.tag
```diff
@@ -246,47 +173,23 @@
                 self.competition.is_admin = CODALAB.state.user.has_competition_admin_privileges(competition)
                 self.update()
                 $('.tabular.menu .item').tab();
+                _.forEach(competition.pages, (page, index) => {
+                    $(`#page_${index}`)[0].innerHTML = sanitize_HTML(EasyMDE.prototype.markdown(page.content))
```
ckcollab: Does this ever have race conditions where it tries to set something that doesn't exist?

src/apps/competitions/models.py
```diff
@@ -85,6 +85,9 @@ class Phase(models.Model):
     solutions = models.ManyToManyField('tasks.Solution', blank=True, related_name="phases")
     is_task_and_solution = models.BooleanField(default=False)
 
+    class Meta:
+        ordering = ('index',)
```
ckcollab: kewl 👍

src/static/riot/competitions/detail/_tabs.tag
jimmykodes: Not that I've seen, but I suppose it could? Are you thinking like trying to set `inner_html` before the `self.update()` has actually created the `<div>`s?

ckcollab: > Are you thinking like trying to set inner_html before the self.update() has actually created the <div>s?

Something like that, yeah, if you haven't ever experienced it then it's probably not a huge deal

jimmykodes:
Tests are passing now, How we looking?

PR #196, opened 2019-05-21
Remove Datasets from Phases

jimmykodes:
Shift to just using tasks for phases.


src/apps/api/tests/test_phases.py
```diff
@@ -41,7 +41,7 @@ def test_collab_can_re_run_whole_phase(self):
         self.client.login(username='collab', password='test')
         self.assert_rerun_succeeds()
 
-    def test_normal_user_can_re_run_whole_phase(self):
+    def test_normal_user_cannot_re_run_whole_phase(self):
```
ckcollab: whoops...!

src/apps/competitions/tasks.py
```diff
@@ -116,17 +117,18 @@ def create_detailed_output_file(detail_name, submission):
...
     )
-    submission = Submission.objects.select_related(*related_models).prefetch_related('details').get(pk=submission_pk)
+    prefetch_models = (
+        'details',
+        'phase__tasks__input_data',
+        'phase__tasks__reference_data',
+        'phase__tasks__scoring_program',
+        'phase__tasks__ingestion_program',
+    )
+    submission = Submission.objects.select_related(*select_models).prefetch_related(*prefetch_models).get(pk=submission_pk)
```
ckcollab: Hmm, I think this is OK the way it is, what do you think about this though? Will leave it up to you what style is better
```python
qs = Submission.objects.select_related(*select_models).prefetch_related(*prefetch_models)
submission = qs.get(pk=submission_pk)
```

src/apps/competitions/tasks.py
```diff
@@ -217,14 +192,22 @@ def get_data_key(obj, file_type, temp_directory, creator):
         return new_dataset.key
     elif len(file_name) in (32, 36):
         # UUID are 32 or 36 characters long
-        # TODO send error message if invalid UUID or invalid filename
-        # if filename is 32 or 36 chars long but isn't present,
-        # it processes like UUID and then breaks but doesn't inform user.
+        if not Data.objects.filter(key=file_name).exists():
+            raise CompetitionUnpackingException(f'Cannot find {file_type} with key: "{file_name}" for task: "{obj["name"]}"')
```
ckcollab: ty

src/apps/competitions/tasks.py
```diff
@@ -247,7 +230,7 @@ def unpack_competition(competition_dataset_pk):
 
     # Children datasets are those that are created specifically for this "parent" competition.
     # They will be deleted if the competition creation fails
-    children_datasets = []
+    # children_datasets = []
```
ckcollab: pull?

src/apps/competitions/tasks.py
```diff
@@ -384,16 +366,17 @@ def unpack_competition(competition_dataset_pk):
...
+                        solution.tasks.add(*Task.objects.filter(key__in=task_keys))
 
                         competition['solutions'][index] = solution['key']
 
                     else:
                         # create solution object and then add {index: {'key': key, 'tasks': task_indexes}} to competition solutions
-                        name = solution[
-                            'name'] if 'name' in solution else f"solution @ {now().strftime('%m-%d-%Y %H:%M')}"
-                        description = solution['description'] if 'description' in solution else None
+                        name = solution.get('name') or f"solution @ {now().strftime('%m-%d-%Y %H:%M')}"
```
ckcollab: Can do f string formatting of dates like:

```python
f"solution @ {now():%m-%d-%Y %H:%M}"
```

src/apps/competitions/tasks.py
```diff
@@ -423,65 +406,29 @@ def unpack_competition(competition_dataset_pk):
...
+                    "description": phase_data.get('description'),
+                    "start": _get_datetime(phase_data.get('start')),
+                    "end": _get_datetime(phase_data.get('end')),
+                    'max_submissions_per_day': phase_data.get('max_submissions_per_day'),
+                    'max_submissions_per_person': phase_data.get('max_submissions'),
                 }
-                if 'execution_time_limit_ms' in phase_data:
-                    new_phase['execution_time_limit'] = phase_data.get('execution_time_limit_ms')
-                if 'max_submissions_per_day' in phase_data or 'max_submissions' in phase_data:
+                execution_time_limit = phase_data.get('execution_time_limit_ms')
```
ckcollab: Hmm, should we add `_ms` or remove it to make these the same on both ends?

src/apps/competitions/tasks.py
jimmykodes: Yea i like your styling better. doing that.

src/apps/competitions/tasks.py
jimmykodes: brilliant!

src/apps/competitions/tasks.py
jimmykodes: I think having `_ms` on the yaml is a good idea so it is explicit to the users what units the time limit is in. making the field the same name on the backend would probably be nice

src/apps/competitions/tasks.py
ckcollab: Python is so pretty :)

src/apps/competitions/tasks.py
jimmykodes: I was leaving it as a placeholder/reminder to eventually actually implement that. I can pull it, or mark it with a TODO, or whatever makes the most sense.

ckcollab: `# TODO: ` would be fine

src/apps/competitions/tasks.py
jimmykodes: I think we should leave this how it is for now since going through and changing all of that is outside the scope of this PR.

ckcollab: Sounds good, I'll add a Trello item for it

jimmykodes:
@ckcollab any other changes to make?

ckcollab:
:shipit:

PR #202, opened 2019-06-11
197 phase migration

### LINKED ISSUE #197: Add Phase-to-phase migration and auto-migration between phases
- [x] Add `auto_migration` option to phases. When phase starts it grabs the submissions from the previous phase and executes them

- [x] Code + tests exist in v1.5 for this should be easy to port

BrendanMyers88:
Resolves #197 

- [x] Adds an auto-migration for phases when new phase begins and auto-migration is enabled for that phase.

- [x] Don't allow `auto_migration` flag on first phase, because `auto_migration` is for the phase being migrated to. 
- [x] Don't allow `auto_migration` if your competition only has one phase

- [x] Setup a scheduled job that checks every 5 minutes for phases that need migrating

**Tests written for...**
- [x] Migrating Phases
- [x] Competition does not migrate when it's already migrating
- [x] Competition does not migrate when it is scoring submissions
- [x] Competitions auto-migrate
- [x] Submissions are run after migrating


src/apps/competitions/models.py
```diff
@@ -23,6 +28,107 @@ def __str__(self):
...
+            final_phase = phases.get(status='Final')
+
+        except Phase.DoesNotExist:
+            return
+
+        # Making sure current_phase or next_phase is not None
+        if current_phase is None or next_phase is None or current_phase is final_phase:
+            return
+
+        logger.info("Checking for needed migrations on competition pk=%s, current phase: %s, next phase: %s" %
```
ckcollab: One thing that is annoying on v1.5 is I see these messages all the time for competitions already migrated. Would be nice if these competitions _never_ ran the check again... some kind of `has_been_migrated` flag?

src/apps/competitions/models.py
```diff
@@ -23,6 +28,107 @@ def __str__(self):
...
+    def apply_phase_migration(self, current_phase, next_phase):
+        '''
+        Does the actual migrating of submissions from current_phase to next_phase
+
+        :param current_phase: The phase object to transfer submissions from
+        :param next_phase: The new phase object we are entering
+        '''
+        logger.info("Checking for submissions that may still be running competition pk=%s" % self.pk)
+
+        if current_phase.submissions.filter(status='Scoring').exists():
```
ckcollab: The status check here should include more statuses, here are all the options currently:

```
    CANCELLED = "Cancelled"
    FINISHED = "Finished"
    FAILED = "Failed"
```

I think may be best to instead have a `completed_when` field on submissions? (what do you think about that naming, along with `created_when`...) I made a new issue for this here https://github.com/codalab/competitions-v2/issues/203

For now check if there are any submissions NOT "finished, failed, or cancelled"

src/apps/competitions/models.py
```diff
@@ -23,6 +28,107 @@ def __str__(self):
...
+            participants = {}
+
+            for s in submission_list:
+                if s.is_migrated is False:
+                    participants[s.participant] = s
+
+            for participant, submission in participants.items():
+                logger.info('Moving submission %s over' % submission)
+
+                file_args = {
```
ckcollab: I don't think we need to do this any more? Can send along normally to below `Submission`?

src/apps/competitions/models.py
```diff
@@ -23,6 +28,107 @@ def __str__(self):
...
+                submission.save()
+
+        except Submission.DoesNotExist:
+            pass
+
+        # To check for submissions being migrated, does not allow to enter new submission
+        next_phase.is_migrated = True
+        next_phase.save()
+
+        # TODO: ONLY IF SUCCESSFUL
```
ckcollab: What about this todo?

src/apps/competitions/models.py
```diff
@@ -101,6 +208,14 @@ def can_user_make_submissions(self, user):
                 return False, 'Reached maximum allowed submissions for this phase'
         return True, None
 
+    @property
+    def is_active(self):
```
ckcollab: cool.

src/apps/competitions/tests/test_phase_migration.py
```diff
@@ -0,0 +1,98 @@
...
+        kwargs.setdefault('phase', self.phase1)
+        kwargs.setdefault('status', 'None')
+        sub = SubmissionFactory(**kwargs)
+        SubmissionScoreFactory(submission=sub)
+        return sub
+
+    def test_migrate_submissions(self):
+        """
+        Determine whether all submissions in phase1 end up in phase2
+        :return:
```
ckcollab: remove the returns here, I believe proper style is to do docstrings like so:

```python
"""I'm on the same line as the quotes!"""
```

..but, I think these are probably unnecessary. Test names should be of the form: `test_<what i am testing>_<what i expect>` like `test_migrating_submissions_moves_submissions_to_next_phase`

src/apps/competitions/models.py
```diff
@@ -23,6 +28,107 @@ def __str__(self):
...
+                self.apply_phase_migration(current_phase, next_phase)
+
+    def apply_phase_migration(self, current_phase, next_phase):
+        '''
+        Does the actual migrating of submissions from current_phase to next_phase
+
+        :param current_phase: The phase object to transfer submissions from
+        :param next_phase: The new phase object we are entering
+        '''
+        logger.info("Checking for submissions that may still be running competition pk=%s" % self.pk)
```
jimmykodes: Let's update all of there to be f strings.

src/apps/competitions/models.py
```diff
@@ -23,6 +28,107 @@ def __str__(self):
...
+            self.is_migrating_delayed = True
+            self.save()
+            return
+        else:
+            logger.info("No submissions running for competition pk=%s" % self.pk)
+
+        logger.info('Doing phase migration on competition pk=%s from phase: %s to phase: %s' %
+                    (self.pk, current_phase.index, next_phase.index))
+
+        if self.is_migrating:
```
jimmykodes: Check this before we look for submissions. That way we can potentially save some unnecessary database queries.

src/apps/competitions/models.py
```diff
@@ -23,6 +28,107 @@ def __str__(self):
...
+            submissions = Submission.objects.filter(phase=current_phase)
+
+            for submission in submissions:
+                submission_list.append(submission)
+
+            participants = {}
+
+            for s in submission_list:
+                if s.is_migrated is False:
+                    participants[s.participant] = s
```
jimmykodes: If I understand this logic right, this only migrates one submission per user each time this runs, because the participant object is the same for all the user's submissions, so it just keeps overwriting the same key. Is this the desired functionality?

src/apps/competitions/tests/test_phase_migration.py
```diff
@@ -0,0 +1,98 @@
...
+        self.competition_participant2 = CompetitionParticipantFactory(user=self.normal_user2,
+                                                                      competition=self.competition)
+        self.competition_participant3 = CompetitionParticipantFactory(user=self.normal_user3,
+                                                                      competition=self.competition)
+        self.phase1 = PhaseFactory(competition=self.competition, auto_migration=True, status='Current')
+        self.phase2 = PhaseFactory(competition=self.competition, auto_migration=True, status='Next')
+        self.phase3 = PhaseFactory(competition=self.competition, auto_migration=False, status='Final')
+        self.submission = self.make_submission()
+        self.submission2 = self.make_submission(participant=self.competition_participant2)
+        self.client = APIClient()
```
jimmykodes: unnecessary setup since you aren't calling `self.client.login()` or anything of the like anywhere

src/apps/competitions/tests/test_phase_migration.py
```diff
@@ -0,0 +1,98 @@
...
+from django.test import TestCase
+from django.contrib.auth import get_user_model
+from django.utils.timezone import now
+from rest_framework.test import APIClient
+
+from competitions.tasks import do_phase_migrations
+from factories import UserFactory, CompetitionFactory, PhaseFactory, SubmissionFactory, SubmissionScoreFactory, \
+    CompetitionParticipantFactory
+
+User = get_user_model()
```
jimmykodes: Not using this anywhere?

src/apps/competitions/tests/test_phase_migration.py
```diff
@@ -0,0 +1,98 @@
...
+        self.submission2 = self.make_submission(participant=self.competition_participant2)
+        self.client = APIClient()
+
+    def make_submission(self, **kwargs):
+        kwargs.setdefault('owner', self.owner)
+        kwargs.setdefault('participant', self.competition_participant)
+        kwargs.setdefault('phase', self.phase1)
+        kwargs.setdefault('status', 'None')
+        sub = SubmissionFactory(**kwargs)
+        SubmissionScoreFactory(submission=sub)
```
jimmykodes: Can probably remove this, since we don't need submission scores for anything in these tests.

src/apps/competitions/tests/test_phase_migration.py
```diff
@@ -0,0 +1,98 @@
...
+    def test_migrating_scoring_submissions(self):
+        """
+        Test to see if there are submissions being scored currently. If there are, delay migration.
+        :return:
+        """
+        self.submission3 = self.make_submission(status='Scoring', participant=self.competition_participant3)
+        self.competition.apply_phase_migration(self.phase1, self.phase2)
+        self.assertFalse(self.phase1.submissions.count() == self.phase2.submissions.count())
+
+    def test_auto_migrate(self):
```
jimmykodes: We should make another test that makes sure we don't auto migrate if the phase isn't supposed to auto migrate. Like maybe modify phase 2 to `auto_migrate=False` for a test and make sure we migrate from 1 -> 2 but not 2 -> 3.

src/apps/competitions/models.py
```diff
@@ -67,6 +173,7 @@ class Phase(models.Model):
     name = models.CharField(max_length=256)
     description = models.TextField(null=True, blank=True)
     execution_time_limit = models.PositiveIntegerField(default=60 * 10)
+    auto_migration = models.BooleanField(default=False)
```
jimmykodes: We may want to rename this to be a little more clear, are we auto migrating _from_ this phase or _to_ this phase. I image from, but clarity in naming could be helpful, especially in potentially creating competitions where they only want to migrate certain phases of the comp.

Also, the competition unpacking task should be updated to look for auto migration in phases on the yaml.

src/apps/competitions/models.py
jimmykodes: can we do something like: 
```python
for submission in Submission.objects.filter(phase=current_phase, is_migrated=False):
    new_submission = Submission(
        participant=submission.participant,
        # and all the other fields
    )
    new_submission.save(ignore_submission_limit=True)
    new_submission.start()

    submission.is_migrated = True
    submission.save()
```
Then we do this more than one submission per participant at a time. (unless that is the desired functionality, in which case I have some ideas for cool nested queries so you can still just loop a single queryset)

src/apps/competitions/models.py
```diff
@@ -23,6 +28,107 @@ def __str__(self):
...
+        # Making sure current_phase or next_phase is not None
+        if current_phase is None or next_phase is None or current_phase is final_phase:
+            return
+
+        logger.info("Checking for needed migrations on competition pk=%s, current phase: %s, next phase: %s" %
+                    (self.pk, current_phase.index, next_phase.index))
+
+        # Check for next phase and see if it has auto_migration enabled
+        if next_phase and next_phase.auto_migration:
+            if next_phase.start < now() and current_phase.end < now():
```
jimmykodes: Probably also `and current_phase.auto_migration` so we don't migrate if we aren't supposed to.

ckcollab:
Btw added these two checkboxes to the description:
```
- [ ] Don't allow `auto_migration` flag on first phase, because `auto_migration` is for the phase being migrated to. 
- [ ] Don't allow `auto_migration` if your competition only has one phase```

File: src/apps/competitions/tasks.py [hunk 16]
```diff
@@ -680,3 +680,11 @@ def create_competition_dump(competition_pk, keys_instead_of_files=True):
         logger.info(f"Finished creating competition dump: {temp_comp_dump.pk} for competition: {comp.pk}")
     except ObjectDoesNotExist:
         logger.info("Could not find competition with pk {} to create a competition dump".format(competition_pk))
+
+
+@app.task(queue='site-worker', soft_time_limit=60 * 5)
+def do_phase_migrations():
+    competitions = Competition.objects.filter(is_migrating=False)
```
jimmykodes: I'm not 100% sure this will work as intended, but consider:
`competitions = Competition.objects.filter(is_migrating=False, phases__auto_migration=True)`
This should be tested, but I _think_ this should filter out any competitions that don't have any phases with `auto_migration=True`, will save some work for the site_worker if that is true. If this doesn't work, I'm pretty damn sure we could come up with a query that _will_ do that.

src/apps/competitions/models.py
BrendanMyers88: Resolved in latest push

src/apps/competitions/models.py
BrendanMyers88: Resolved in latest push

src/apps/competitions/tests/test_phase_migration.py
BrendanMyers88: Cruft removed in latest push and made test names clearer for their objective.

src/apps/competitions/models.py
BrendanMyers88: Updated these to be f strings in latest push. Still lots of conversion to do on v2 as a whole though.

src/apps/competitions/models.py
BrendanMyers88: Added a more accurate filter to prevent unnecessary competition database queries.

src/apps/competitions/models.py
BrendanMyers88: replaced the logic with what you've got and it looks much cleaner! Thanks Jimmy!

src/apps/competitions/tests/test_phase_migration.py
BrendanMyers88: Cruft removed in latest push

src/apps/competitions/tests/test_phase_migration.py
BrendanMyers88: Cruft removed in latest push

src/apps/competitions/tests/test_phase_migration.py
BrendanMyers88: Cruft removed in latest push

src/apps/competitions/tests/test_phase_migration.py
BrendanMyers88: Named variables more clearly by naming it  `auto_migrate_to_this_phase`. Also added a test where we try to push to an `auto_migrate_to_this_phase = False` in phase 3 after `auto_migrate_to_this_phase = True` in phase 2.

src/apps/competitions/models.py
BrendanMyers88: See above, variable name is clearer now: `auto_migrate_to_this_phase`

src/apps/competitions/tasks.py
BrendanMyers88: Added this logic and this cut down on db calls. I didn't know about the `<model>__<model_field>` so that was really helpful. Thanks!

src/apps/competitions/migrations/0035_auto_20190612_2052.py
```diff
@@ -0,0 +1,27 @@
...
+        migrations.RemoveField(
+            model_name='competition',
+            name='has_been_migrated',
+        ),
+        migrations.AddField(
+            model_name='phase',
+            name='has_been_migrated',
+            field=models.BooleanField(default=False),
+        ),
+    ]
```
jimmykodes: Squash these migrations to as few as required.

src/apps/competitions/models.py
```diff
@@ -6,23 +7,75 @@
...
+    def apply_phase_migration(self, current_phase, next_phase):
+        '''
+        Does the actual migrating of submissions from current_phase to next_phase
+
+        :param current_phase: The phase object to transfer submissions from
+        :param next_phase: The new phase object we are entering
+        '''
+        logger.info(f"Checking for submissions that may still be running competition pk={self.pk}")
+
+        if not current_phase.submissions.filter(status='Scoring' or 'Cancelled' or 'Finished').exists():
```
jimmykodes: I think this is the correct logic for this query:
```python
if not current_phase.submissions.exclude(status__in=['Scoring', 'Cancelled', 'Finished']).exists()
```

src/apps/competitions/tests/test_phase_migration.py
```diff
@@ -0,0 +1,106 @@
...
+        kwargs.setdefault('participant', self.competition_participant)
+        kwargs.setdefault('phase', self.phase1)
+        kwargs.setdefault('status', 'None')
+        sub = SubmissionFactory(**kwargs)
+        SubmissionScoreFactory(submission=sub)
+        return sub
+
+    def test_migrate_submissions(self):
+        assert self.phase2.submissions.count() == 0
+        do_phase_migrations.apply()
```
jimmykodes: Do we need to run these as tasks, or can we just call the function directly? `do_phase_migrations()`

src/apps/competitions/tests/test_phase_migration.py
```diff
@@ -0,0 +1,106 @@
...
+        assert self.phase1.submissions.count() == self.phase2.submissions.count()
+
+    def test_currently_migrating_competitions_dont_migrate(self):
+        self.competition.is_migrating = True
+        self.competition.save()
+        do_phase_migrations.apply()
+        self.assertFalse(self.phase1.submissions.count() == self.phase2.submissions.count())
```
jimmykodes: generally prefer plain assert statements: 
```python
assert not self.phase1.submissions.count() == self.phase2.submissions.count()
# or
assert self.phase1.submissions.count() != self.phase2.submissions.count()
```

src/apps/competitions/tests/test_phase_migration.py
```diff
@@ -0,0 +1,106 @@
...
+
+        # Competition 1 Submissions
+        self.submission = self.make_submission()
+        self.submission2 = self.make_submission(participant=self.competition_participant2)
+
+        # Competition 2 Submissions
+        self.submission1_competition2 = self.make_submission(phase=self.phase1_competition2,
+                                                             participant=self.competition_participant)
+        self.submission2_competition2 = self.make_submission(phase=self.phase1_competition2,
+                                                             participant=self.competition_participant3)
```
jimmykodes: Do we need all this setup for all of these tests? can we pare this down into only what we need for every test scenario and then supplement in the tests themselves?

src/apps/competitions/tests/test_phase_migration.py
```diff
@@ -0,0 +1,106 @@
...
+
+from django.test import TestCase
+from django.utils.timezone import now
+
+from competitions.tasks import do_phase_migrations
+from factories import UserFactory, CompetitionFactory, PhaseFactory, SubmissionFactory, SubmissionScoreFactory, \
+    CompetitionParticipantFactory
+
+twenty_minutes_ago = now() - datetime.timedelta(hours=0, minutes=20)
+five_minutes_ago = now() - datetime.timedelta(hours=0, minutes=5)
```
jimmykodes: Probably want these in the setup so they stay relative to the tests themselves, rather than fixed for all the tests?

src/apps/competitions/tests/test_phase_migration.py
```diff
@@ -0,0 +1,106 @@
...
+        self.assertFalse(self.phase1.submissions.count() == self.phase2.submissions.count())
+
+    def test_competitions_with_scoring_submissions_dont_migrate(self):
+        self.submission3 = self.make_submission(status='Scoring', participant=self.competition_participant3)
+        do_phase_migrations.apply()
+        self.assertFalse(self.phase1.submissions.count() == self.phase2.submissions.count())
+
+    def test_auto_migration_at_new_phase(self):
+        do_phase_migrations.apply()
+        assert self.phase1.submissions.count() == self.phase2.submissions.count()
```
jimmykodes: This appears to be a duplicate to `test_migrate_submissions`

src/apps/competitions/tests/test_phase_migration.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def test_submission_ran_after_migration_complete(self):
+        do_phase_migrations.apply()
+        for submission in self.phase2.submissions.all():
```
jimmykodes: `assert not self.phase2.submissions.filter(status='None').exists()`

src/apps/competitions/tests/test_phase_migration.py
```diff
@@ -0,0 +1,106 @@
...
+            assert (submission.status is not 'None')
+
+    def test_has_been_migrated_competitions_arent_migrated(self):
+        assert self.phase2_competition2.submissions.count() == 0
```
jimmykodes: `assert not self.phase2_competition2.submissions.exists()`
(might require `submissions.all().exists()` but I don't think so)

src/apps/competitions/tests/test_phase_migration.py
```diff
@@ -0,0 +1,106 @@
...
+        do_phase_migrations.apply()
+        self.assertFalse(self.phase1_competition2.submissions.count() == self.phase2_competition2.submissions.count())
+
+    def test_prevent_migration_to_auto_migrate_to_the_phase_is_false(self):
+        assert self.phase2.submissions.count() == 0
```
jimmykodes: same as above `submissions.exists()`

src/apps/competitions/migrations/0035_auto_20190612_2052.py
BrendanMyers88: May do this in the future, going to hold off for now

src/apps/competitions/models.py
BrendanMyers88: Used Scoring.CANCELLED, etc. and added NONE as well to prevent typos from causing problems. Corrected the Scoring status logger

src/apps/competitions/tests/test_phase_migration.py
BrendanMyers88: Fixed in latest push

src/apps/competitions/tests/test_phase_migration.py
BrendanMyers88: Switched all `self.assert`s to `assert` or `assert not` in latest push

src/apps/competitions/tests/test_phase_migration.py
BrendanMyers88: Removed a lot of unused setup and added a loop for submissions in the setup.

src/apps/competitions/tests/test_phase_migration.py
BrendanMyers88: pulled duplicate test

src/apps/competitions/tests/test_phase_migration.py
BrendanMyers88: Changed from loop to filter in latest push

src/apps/competitions/tests/test_phase_migration.py
BrendanMyers88: Much cleaner way to do this, thanks Jimmy!

src/apps/api/serializers/competitions.py
```diff
@@ -98,6 +100,15 @@ def validate_leaderboards(self, value):
             raise serializers.ValidationError("Competitions require at least 1 leaderboard")
         return value
 
+    def validate_phases(self, phases):
+        if not phases or len(phases) <= 0:
+            raise serializers.ValidationError("Competitions must have at least one phase")
+        if phases[0]['auto_migrate_to_this_phase'] is True:
+            raise serializers.ValidationError("You cannot auto migrate to the first phase of a competition")
+        if len(phases) == 1 and phases[0]['auto_migrate_to_this_phase']:
```
jimmykodes: This validation step is unreachable. Need to move it above checking just `phase[0]` in order for it to trigger.

src/apps/api/tests/test_phases.py
```diff
@@ -46,3 +49,42 @@ def test_normal_user_cannot_re_run_whole_phase(self):
...
+
+    def serialize_and_validate_data(self):
+        serializer = CompetitionSerializer(self.competition)
+        data = serializer.data
+
+        with pytest.raises(ValidationError) as exception:
+            serializer = CompetitionSerializer(data=data)
+            serializer.is_valid(raise_exception=True)
+
+        return self, exception
```
jimmykodes: You can probably just return the exception here, since you aren't using the `self` anywhere

src/apps/api/tests/test_phases.py
```diff
@@ -46,3 +49,42 @@ def test_normal_user_cannot_re_run_whole_phase(self):
...
+
+    def test_phase_is_valid(self):
+        setattr(self.phase, 'auto_migrate_to_this_phase', False)
+        self.phase.save()
+        _, exception = self.serialize_and_validate_data()
+
+        self.assertFalse("'phase:'" in str(exception.value))
```
jimmykodes: Can we restructure this test so that we can assert that an exception wasn't raised at all? Is that even doable in a test environment, or does the validator require some data that can't be "factoried" easily?

src/apps/api/tests/test_phases.py
```diff
@@ -46,3 +49,42 @@ def test_normal_user_cannot_re_run_whole_phase(self):
...
+        data = serializer.data
+
+        with pytest.raises(ValidationError) as exception:
+            serializer = CompetitionSerializer(data=data)
+            serializer.is_valid(raise_exception=True)
+
+        return self, exception
+
+    def test_phase_is_valid(self):
+        setattr(self.phase, 'auto_migrate_to_this_phase', False)
```
jimmykodes: Should just be able to do 
```python
self.phase.auto_migrate_to_this_phase = False
self.phase.save()
```

src/apps/competitions/models.py
```diff
@@ -6,23 +7,75 @@
...
+        self.is_migrating = True
+        self.save()
+
+        submissions = Submission.objects.filter(phase=current_phase, is_migrated=False)
+
+        for submission in submissions:
+            new_submission = Submission(
+                participant=submission.participant,
+                phase=next_phase,
+                result=submission.result,
```
jimmykodes: Should be able to pull this field.

src/apps/competitions/models.py
jimmykodes: We could add a field to phases that is something like `watch_list` or something like that which saves the pks of the submissions it created, then in the celery job that looks for phases that need to be migrated, also run
```python
Phase.objects.annotate(running_subs=Count(Subquery(Submission.objects.filter(pk__in=OuterRef('watch_list')).exclude(
    status__in=['Finished', 'Failed', 'Cancelled']).values('pk')))).filter(running_subs=0, is_migrating=True).update(
    is_migrating=False)
```
Pretty sure this will update phases to `is_migrate=False` if their `watch_list` submissions are in a "done" status. Should do it all in one query too ;) 

@ckcollab Thoughts on this as a solution to this todo?

src/apps/competitions/tasks.py
```diff
@@ -680,3 +680,13 @@ def create_competition_dump(competition_pk, keys_instead_of_files=True):
         logger.info(f"Finished creating competition dump: {temp_comp_dump.pk} for competition: {comp.pk}")
     except ObjectDoesNotExist:
         logger.info("Could not find competition with pk {} to create a competition dump".format(competition_pk))
+
+
+@app.task(queue='site-worker', soft_time_limit=60 * 5)
+def do_phase_migrations():
+    new_phases = Phase.objects.filter(auto_migrate_to_this_phase=True, start__lte=now(),
```
jimmykodes: should we do
```python
start__lte=now() + timedelta(minutes=30)
```
and start migrating a little before the next phase actually starts?

src/apps/competitions/tests/test_phase_migration.py
```diff
@@ -0,0 +1,74 @@
...
+    def test_migrate_submissions(self):
+        assert not self.phase2.submissions.exists()
+        do_phase_migrations()
+        assert self.phase1.submissions.count() == self.phase2.submissions.count()
+
+    def test_currently_migrating_competitions_dont_migrate(self):
+        self.competition.is_migrating = True
+        self.competition.save()
+        do_phase_migrations()
+        assert not self.phase1.submissions.count() == self.phase2.submissions.count()
```
jimmykodes: another small nitpick, `assert x != y` rather than `assert not x == y`

src/apps/api/tests/test_phases.py
```diff
@@ -46,3 +49,43 @@ def test_normal_user_cannot_re_run_whole_phase(self):
...
+
+        assert ("'phase:'" not in str(exception.value))
+
+    def test_phase_serializer_auto_migrate_on_first_phase(self):
+        self.phase.auto_migrate_to_this_phase = True
+        self.phase.save()
+        exception = self.serialize_and_validate_data()
+
+        assert ("You cannot auto migrate in a competition with one phase" or
+                "You cannot auto migrate to the first phase of a competition" in str(exception.value))
```
jimmykodes: This is the equivalent of
```python
assert "You cannot auto migrate in a competition with one phase" in str(exception.value))
```
for the logic you want, either:
```python
assert "You cannot auto migrate in a competition with one phase" in str(exception.value) \
       or "You cannot auto migrate to the first phase of a competition" in str(exception.value)
```
or
```python
errors = ["You cannot auto migrate in a competition with one phase", "You cannot auto migrate to the first phase of a competition"]
assert any(error in str(exception.value) for error in errors)

```

src/apps/api/serializers/competitions.py
BrendanMyers88: Made code reachable in latest push

src/apps/api/tests/test_phases.py
BrendanMyers88: Function now returns only exception

src/apps/api/tests/test_phases.py
BrendanMyers88: Switched to bare asserts for testing. I wasn't able to figure out how to do this in the test environment but maybe we can attack this together?

src/apps/api/tests/test_phases.py
BrendanMyers88: Made attribute changes more readable

src/apps/competitions/models.py
BrendanMyers88: Pulled field in latest push

src/apps/competitions/tasks.py
BrendanMyers88: I think this would prevent scoring for submissions made in the final 30 minutes of the phase, but not positive.

src/apps/competitions/tests/test_phase_migration.py
BrendanMyers88: Changed all `assert not x == y` to `assert x !=y`

src/apps/api/tests/test_phases.py
BrendanMyers88: Ah right! I make this mistake too often, thanks Jimmy!

src/apps/competitions/tasks.py
jimmykodes: Ah. Pretty sure you are right.

PR #205, opened 2019-06-17
Competition Dump Button Location Change

jimmykodes:
added create dump button on competition detail page and a file table for bundles/dumps
image

### DIFF
--- a/Dockerfile.builder
+++ b/Dockerfile.builder
@@ -7,6 +7,4 @@ WORKDIR /app
 # Install packages
 ADD package.json .
 
-RUN npm install
-
-CMD export PATH=./node_modules/.bin:$PATH && npm-watch
+CMD npm install && export PATH=./node_modules/.bin:$PATH && npm-watch
--- a/src/apps/api/views/competitions.py
+++ b/src/apps/api/views/competitions.py
@@ -7,6 +7,7 @@ from rest_framework.viewsets import ModelViewSet, GenericViewSet
 from api.serializers.competitions import CompetitionSerializer, CompetitionSerializerSimple, PhaseSerializer, \
     CompetitionCreationTaskStatusSerializer, CompetitionDetailSerializer
 from competitions.models import Competition, Phase, CompetitionCreationTaskStatus
+from utils.data import make_url_sassy
 
 
 class CompetitionViewSet(ModelViewSet):
@@ -74,22 +75,24 @@ class CompetitionViewSet(ModelViewSet):
 
     @action(detail=True, methods=('GET',))
     def get_files(self, request, pk):
-        competition = Competition.objects.select_related(
+        qs_helper = Competition.objects.select_related(
             'created_by'
         ).prefetch_related(
             'dumps__dataset',
-        ).get(id=pk)
+        )
+        competition = qs_helper.get(id=pk)
         if request.user != competition.created_by and request.user not in competition.collaborators.all():
             raise PermissionDenied("You don't have access to the competition files")
-        from utils.data import make_url_sassy
         bundle = competition.bundle_dataset
         files = {
             'bundle': {
                 'name': bundle.name,
                 'url': make_url_sassy(bundle.data_file.name)
             },
-            'dumps': [{'name': dump.dataset.name, 'url': make_url_sassy(dump.dataset.data_file.name)} for dump in competition.dumps.all()]
+            'dumps': []
         }
+        for dump in competition.dumps.all():
+            files['dumps'].append({'name': dump.dataset.name, 'url': make_url_sassy(dump.dataset.data_file.name)})
         return Response(files)
 
 
--- a/src/apps/api/views/competitions.py
+++ b/src/apps/api/views/competitions.py
@@ -72,6 +72,26 @@ class CompetitionViewSet(ModelViewSet):
         competition.save()
         return Response({"published": competition.published})
 
+    @action(detail=True, methods=('GET',))
+    def get_files(self, request, pk):
+        competition = Competition.objects.select_related(
+            'created_by'
+        ).prefetch_related(
+            'dumps__dataset',
+        ).get(id=pk)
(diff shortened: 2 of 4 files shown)


Dockerfile.builder
```diff
@@ -7,6 +7,4 @@ WORKDIR /app
 # Install packages
 ADD package.json .
 
-RUN npm install
-
-CMD export PATH=./node_modules/.bin:$PATH && npm-watch
+CMD npm install && export PATH=./node_modules/.bin:$PATH && npm-watch
```
ckcollab: Why all of this in the cmd?

src/apps/api/views/competitions.py
```diff
@@ -72,6 +72,26 @@ def toggle_publish(self, request, pk):
...
+        if request.user != competition.created_by and request.user not in competition.collaborators.all():
+            raise PermissionDenied("You don't have access to the competition files")
+        from utils.data import make_url_sassy
+        bundle = competition.bundle_dataset
+        files = {
+            'bundle': {
+                'name': bundle.name,
+                'url': make_url_sassy(bundle.data_file.name)
+            },
+            'dumps': [{'name': dump.dataset.name, 'url': make_url_sassy(dump.dataset.data_file.name)} for dump in competition.dumps.all()]
```
ckcollab: indentation here?

src/apps/api/views/competitions.py
```diff
@@ -72,6 +72,26 @@ def toggle_publish(self, request, pk):
         competition.save()
         return Response({"published": competition.published})
 
+    @action(detail=True, methods=('GET',))
+    def get_files(self, request, pk):
+        competition = Competition.objects.select_related(
```
ckcollab: Hmm, this is OK but I wonder if this is nicer?

```python
qs_helper = Competition.objects.select_related('created_by').prefetch_related('dumps__dataset')
competition = qs_helper.get(id=pk)
```

src/apps/api/views/competitions.py
```diff
@@ -72,6 +72,26 @@ def toggle_publish(self, request, pk):
...
+            'created_by'
+        ).prefetch_related(
+            'dumps__dataset',
+        ).get(id=pk)
+        if request.user != competition.created_by and request.user not in competition.collaborators.all():
+            raise PermissionDenied("You don't have access to the competition files")
+        from utils.data import make_url_sassy
```
ckcollab: do we need to inline this?

src/static/riot/competitions/detail/_tabs.tag
```diff
@@ -85,6 +85,8 @@
                               leaderboards="{ competition.leaderboards }"></leaderboards>
             </div>
         </div>
+
+        <!--admin tab-->
```
ckcollab: ty

src/static/js/ours/client.js
```diff
@@ -77,6 +77,9 @@ CODALAB.api = {
     re_run_phase_submissions: function (phase_pk) {
         return CODALAB.api.request('GET', `${URLS.API}phases/${phase_pk}/rerun_submissions/`)
     },
+    get_competition_files: pk => {
```
ckcollab: did we need to add this because it was called directly before, or something?

src/apps/api/views/competitions.py
jimmykodes: ¯\_(ツ)_/¯ 
Still a little fuzzy on when to do local imports. We are just only using this here, so? I don't know?

src/apps/api/views/competitions.py
jimmykodes: indent how? it is list comprehension, which I never know how to indent in a way that doesn't look confusing to me.

src/apps/api/views/competitions.py
ckcollab: Only when it stops a circular import, do I do something local like that. I'm sure there are a couple other cases, but it's best practice to include imports at the top.

src/static/js/ours/client.js
jimmykodes: because we don't need the files if we aren't admin, so didn't put them on a serializer. plus it allows up to update _just_ the files after creating a dump, rather than reloading the whole competition. Or at least that was my logic.

src/apps/api/views/competitions.py
jimmykodes: Can go either way? I think mine is more straightforward? but maybe that's just cause I wrote it ;) haha

Dockerfile.builder
jimmykodes: something to do with how RUN doesn't persist into the final container, just cached. If I understand things correctly. This was the only way I could get things to work _every_ time, because of caching containers or something I think?

src/apps/api/views/competitions.py
ckcollab: Yeah I misinterpreted what this was doing before, it'd probably be best pulled out as a standalone variable in a boring old loop for clarity sake... but it's fine either way imo, will leave it up to you

```python
dumps = []
for dump in competition.dumps.all()
    dumps.append({'name': d.dataset.name, 'url': make_url_sassy(d.dataset.data_file.name)})
```

Dockerfile.builder
ckcollab: Annoying. Ty for the fix!

src/apps/api/views/competitions.py
ckcollab: :nudges you towards my boring style:

src/static/js/ours/client.js
ckcollab: right, sorry, made this comment before reading your slack msg, ty!!!

jimmykodes:
This does not pull any of the old functionality. That will get addressed when the dataset management/task management merger happens.

jimmykodes:
@ckcollab Any more changes to be made? Or good to merge?

ckcollab:
:shipit:

PR #206, opened 2019-06-20
Chahub tests

jimmykodes:
Finalized Chahub tests and added do_chahub_retries task.


src/apps/chahub/models.py
```diff
@@ -78,37 +85,31 @@ def save(self, force_to_chahub=False, *args, **kwargs):
         if settings.CHAHUB_API_URL:
             is_valid = self.get_chahub_is_valid()
 
-            logger.info("ChaHub :: {}={} is_valid = {}".format(self.__class__.__name__, self.pk, is_valid))
+            logger.info(f"ChaHub :: {self.__class__.__name__}({self.pk}) is_valid = {is_valid}")
```
ckcollab: ty

src/apps/chahub/models.py
```diff
@@ -78,37 +85,31 @@ def save(self, force_to_chahub=False, *args, **kwargs):
...
                         self.chahub_timestamp = timezone.now()
                         self.chahub_data_hash = data_hash
                         self.chahub_needs_retry = False
                     else:
-                        status = resp.status_code if hasattr(resp, 'status_code') else 'N/A'
-                        body = resp.content if hasattr(resp, 'content') else 'N/A'
-                        logger.info("ChaHub :: Error sending to chahub, status={}, body={}".format(status, body))
+                        status = getattr(resp, 'status_code', 'N/A')
+                        body = getattr(resp, 'content', 'N/A')
+                        logger.info(f"ChaHub :: Error sending to chahub, status={status}, body={body}")
```
ckcollab: above is to just say `N/A` instead of `None`?

src/apps/chahub/tasks.py
```diff
@@ -14,12 +18,39 @@ def send_users_to_chahub():
...
+        # This base exception works for HTTP errors, Connection errors, etc.
+        return
+
+    logger.info("ChaHub is online, checking for objects needing to be re-sent to ChaHub")
+    chahub_models = ChaHubSaveMixin.__subclasses__()
+    for model in chahub_models:
+        needs_retry = model.objects.filter(chahub_needs_retry=True)
+
+        if limit:
+            needs_retry = needs_retry[:limit]
```
ckcollab: Cool adding a `limit` here

src/apps/chahub/tests/test_chahub_mixin.py
```diff
@@ -1,109 +1,137 @@
...
+from competitions.models import Submission
+from factories import UserFactory, CompetitionFactory, DataFactory, SubmissionFactory, PhaseFactory, \
+    CompetitionParticipantFactory
+
+
+class ChaHubTestCase(TestCase):
+    def setUp(self):
+        settings.PYTEST_FORCE_CHAHUB = True
+        # set the url to localhost for tests
+        settings.CHAHUB_API_URL = 'localhost/'
```
ckcollab: should include `http(s)`?

src/apps/chahub/tests/test_chahub_mixin.py
```diff
@@ -1,109 +1,137 @@
...
+
+class SubmissionMixinTests(ChaHubTestCase):
+    def setUp(self):
+        self.user = UserFactory()
+        self.comp = CompetitionFactory(published=True)
+        self.participant = CompetitionParticipantFactory(user=self.user, competition=self.comp)
+        self.phase = PhaseFactory(competition=self.comp)
+        self.data = DataFactory()
+        super().setUp()
+        self.submission = SubmissionFactory.build(
```
ckcollab: factories! Yes!

src/apps/chahub/tests/test_chahub_tasks.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def test_do_retries_picks_up_all_expected_items(self):
+        resp = self.mock_retries()
+        # Should grab 5 each Users, Comps, Datasets, Submissions
+        assert resp.call_count == 20
+
+    def test_do_retries_limit_will_limit_number_of_retries(self):
+        resp = self.mock_retries(limit=2)
+        # Should grab 2 each Users, Comps, Datasets, Submissions
+        assert resp.call_count == 8
```
ckcollab: beautiful

src/apps/chahub/tests/test_chahub_utils.py
```diff
@@ -1,46 +1,38 @@
...
+            requests_post_mock.return_value = HttpResponseBase(status=201)
+            requests_post_mock.return_value.content = ''
+            with mock.patch('apps.chahub.utils.requests.put') as requests_put_mock:
+                requests_put_mock.return_value = HttpResponseBase(status=201)
+                requests_put_mock.return_value.content = ''
+                obj.save()
+            return requests_post_mock, requests_put_mock
+
+    def test_send_to_chahub_posts_new_data_and_puts_updated_data(self):
+        post_resp, put_response = self.mock_chahub_save(self.user)
```
ckcollab: why do we need to check post vs put?

src/apps/chahub/models.py
jimmykodes: Pretty much

src/apps/chahub/tests/test_chahub_mixin.py
jimmykodes: This variable could be gibberish and it would still work cause I never actually hit the endpoint. I just need _some_ value there. So I figured localhost so that if i had my mock in the wrong place i would just hit myself rather than someone else. 
(quit hitting yourself, quit hitting yourself.... haha)

src/apps/chahub/tests/test_chahub_utils.py
jimmykodes: basically just making sure the update vs create was being handled 'correctly'? Tyler said something about Chahub needing put instead of post, so I'm testing that it does what it is supposed to when it is supposed to. (with my limited understanding of chahub)

jimmykodes:
@ckcollab Where we at w/ review on this PR?

src/apps/chahub/tests/test_chahub_utils.py
ckcollab: I see, I believe just post _should_ work and is what we expect to be doing (since it does `update_or_create`). @Tthomas63 want to weigh in on this one?

Tthomas63: I forgot exactly where we landed on this, but I believe on Chahub's side we should handle updating/creating through `POST`. This isn't necessarily correct, but because we do bulk operations, we need to handle existing/new competitions in the same batch. (Create or update)

jimmykodes:
This has been updated to always send a post request. also includes a quick catch from tyler about sending empty logo information on a competition. We good to :ship:?

Tthomas63:
@jimmykodes The fix was actually to not send the `producer` key in data for competitions. It's not required to send that data anymore, and it actually causes problems. The empty logo issue is separate. 👍

jimmykodes:
@Tthomas63 Apparently I'm 🙈.. didn't even see that change. fixed it now.

PR #207, opened 2019-06-25
Fix push to chahub

Tthomas63:


src/apps/competitions/models.py
```diff
@@ -28,14 +29,23 @@ def bundle_dataset(self):
     def get_chahub_endpoint(self):
         return "competitions/"
 
+    def clean_chahub_data(self, temp_data):
```
Tthomas63: Not sure if we should pull this into the main Chahub Save Mixin. Pulled it to here from profiles due to the fact logo can still be empty when we add it to the dict with a conditional.

jimmykodes:
Good catch w/ this @Tthomas63.

We can close this PR and I'll handle this in the Chahub tests PR that I have open, so all the chahub changes can get merged in the same go and not have to handle conflicts.

jimmykodes:
Closing this PR and deleting branch since fixes have been included in different branch

PR #214, opened 2019-07-17
134 competition participant

jimmykodes:


jimmykodes:
This is not _complete_ functionality, but it is damn near the mark. Mostly backend shenanigans to worry about, so I figure getting the UI bits merged soon would be ideal so we don't create lots of conflicts w/ competition detail page rework.

ckcollab:
image

Front page tile should show "1" participants here, but is empty?

src/apps/api/serializers/competitions.py
```diff
@@ -121,6 +121,7 @@ class CompetitionDetailSerializer(serializers.ModelSerializer):
     phases = PhaseDetailSerializer(many=True)
     leaderboards = LeaderboardSerializer(many=True)
     collaborators = CollaboratorSerializer(many=True)
+    participant_status = serializers.CharField(read_only=True)
```
ckcollab: should have a `choices=` option here for validation up-and-down the Django stack?

src/apps/api/views/competitions.py
```diff
@@ -164,3 +189,30 @@ class CompetitionCreationTaskStatusViewSet(RetrieveModelMixin, GenericViewSet):
...
+class CompetitionParticipantViewSet(ModelViewSet):
+    queryset = CompetitionParticipant.objects.all()
+    serializer_class = CompetitionParticipantSerializer
+    filter_backends = (DjangoFilterBackend, SearchFilter)
+    filter_fields = ('user__username', 'user__email', 'status', 'competition')
+    search_fields = ('user__username', 'user__email',)
+
+    def get_queryset(self):
+        qs = super().get_queryset()
+        qs = qs.select_related('user').order_by('user__username')
```
ckcollab: nice, was going to say from `CompetitionParticipantSerializer` make sure we do a `select_related` ;)

src/apps/api/views/competitions.py
```diff
@@ -43,6 +45,11 @@ def get_queryset(self):
                 'leaderboards__columns',
                 'collaborators',
             )
+            sub_query = CompetitionParticipant.objects.filter(
+                competition=OuterRef('pk'),
+                user=self.request.user
+            ).values_list('status')[:1]
+            qs = qs.annotate(participant_status=Subquery(sub_query))
```
ckcollab: `sub_query` could be better named `select_participant` or some such?

src/apps/competitions/emails.py
```diff
@@ -0,0 +1,90 @@
+from utils.email import codalab_send_mail
```
ckcollab: I really like the idea of this!! Makes these nice and easy to test if we ever want to just hit the template or some-such

src/apps/api/views/competitions.py
```diff
@@ -164,3 +189,30 @@ class CompetitionCreationTaskStatusViewSet(RetrieveModelMixin, GenericViewSet):
...
+    filter_fields = ('user__username', 'user__email', 'status', 'competition')
+    search_fields = ('user__username', 'user__email',)
+
+    def get_queryset(self):
+        qs = super().get_queryset()
+        qs = qs.select_related('user').order_by('user__username')
+        return qs
+
+    def update(self, request, *args, **kwargs):
+        if request.method == 'PATCH':
```
ckcollab: if we're already in update why do we need to check if `PATCH`?

jimmykodes:
TODO: make `CompetitionSerializerSimple` aware of participant count so competition tile can display this.

ckcollab:
Can maybe make this a separate issue and tackle it later, but attempting to view a competition when I'm not logged in I get redirected to login. Non-logged in folks should be able to view competition (and Google should be able to crawl it)

ckcollab:
Yikes, editing terms as a page feels weird.... maybe we should revise pages into a table on the main competition form?

Below logo, here:
image

Always seems to be blank? Could be because I'm using an old competition or something?
image

src/apps/api/views/competitions.py
jimmykodes: Is update only `PATCH`? Would `PUT` also trigger update?

jimmykodes:
Currently, terms can only be added via the yaml. The competition editor as a whole need some revamping again. We've added a lot of additional fields to models since the last time changes were made.

src/apps/competitions/emails.py
```diff
@@ -0,0 +1,90 @@
...
+            context,
+            **detail
+        )
+
+
+def send_participation_denied_emails(participant):
+    context = {
+        'participant': participant
+    }
+    email_details = [
```
jimmykodes: Split this out. Violate DRY for the sake of code clarity.

src/apps/api/views/competitions.py
```diff
@@ -34,15 +36,24 @@ def get_queryset(self):
...
-            qs = qs.prefetch_related(
-                'phases',
-                'pages',
-                'leaderboards',
-                'leaderboards__columns',
-                'collaborators',
-            )
+        if self.request.method == 'GET':
+            if self.action == 'list':
+                qs = qs.select_related('created_by')
```
ckcollab: can do `pass` here and factor the `qs = qs.select_related('created_by')` out of both conditionals in this if.

could also not do `pass` and `if self.action != 'list':` I believe?

src/apps/api/views/competitions.py
```diff
@@ -80,6 +91,23 @@ def toggle_publish(self, request, pk):
         competition.save()
         return Response({"published": competition.published})
 
+    @action(detail=True, methods=('POST',))
+    def register(self, request, pk):
+        competition = self.get_object()
+        user = request.user
+        participant = CompetitionParticipant.objects.create(competition=competition, user=user)
+        if user in [competition.created_by] + list(competition.collaborators.all()):
```
ckcollab: > `[competition.created_by] + list(competition.collaborators.all()):`

we do this pattern _a lot_ it feels like, should we have `competition.all_organizers` property, or something? Hmm...

src/apps/api/views/competitions.py
```diff
@@ -164,3 +193,30 @@ class CompetitionCreationTaskStatusViewSet(RetrieveModelMixin, GenericViewSet):
...
+
+class CompetitionParticipantViewSet(ModelViewSet):
+    queryset = CompetitionParticipant.objects.all()
+    serializer_class = CompetitionParticipantSerializer
+    filter_backends = (DjangoFilterBackend, SearchFilter)
+    filter_fields = ('user__username', 'user__email', 'status', 'competition')
+    search_fields = ('user__username', 'user__email',)
+
+    def get_queryset(self):
+        qs = super().get_queryset()
```
ckcollab: permissions checks?

src/apps/competitions/emails.py
```diff
@@ -0,0 +1,75 @@
+from utils.email import codalab_send_mail
+
+
+def get_organizer_emails(competition):
```
ckcollab: should be property on `Competition` mebe?

src/apps/competitions/tasks.py
```diff
@@ -274,6 +276,24 @@ def unpack_competition(competition_dataset_pk):
                 "solutions": {},
             }
 
+            # ---------------------------------------------------------------------
+            # Terms
+            if competition['registration_required']:
```
ckcollab: nice!

src/static/js/ours/client.js
```diff
@@ -241,6 +245,15 @@ CODALAB.api = {
     user_lookup: (filters) => {
         return CODALAB.api.request('GET', `${URLS.API}user_lookup/`, filters)
     },
+    /*---------------------------------------------------------------------
+         Participants
+    ---------------------------------------------------------------------*/
+    get_participants: filters => {
+        return CODALAB.api.request('GET', `${URLS.API}participants/`, filters)
+    },
+    update_status: (pk, data) => {
```
ckcollab: `update_participation_status`?

src/static/riot/competitions/detail/_registration.tag
```diff
@@ -0,0 +1,95 @@
+<registration>
+    <div if="{!status}" class="ui grid">
+        <div class="row">
+            <p>You have not yet registered for this competition.</p>
+            <p>
+                To participate in this competition, you must accept its specific <a href="" onclick="{show_modal}">terms and conditions</a>. After you
```
ckcollab: Hmm.. if you remove the href does it still need prevent_default and such wiring, I wonder?

src/static/riot/competitions/detail/_registration.tag
```diff
@@ -0,0 +1,95 @@
...
+
+        self.accept_toggle = () => {
+            self.accepted = !self.accepted
+        }
+
+        self.show_modal = (e) => {
+            if (e) {
+                e.preventDefault()
+            }
+            console.log('hello')
```
ckcollab: pull

src/static/riot/competitions/detail/_registration.tag
```diff
@@ -0,0 +1,95 @@
...
+        </div>
+        <div class="row">
+            <button class="ui primary button {disabled: !accepted}" onclick="{submit_registration}">
+                Register
+            </button>
+        </div>
+    </div>
+
+    <div if="{status}">
+        Your current status is: {_.startCase(status)}
```
ckcollab: lodash is so coool!

src/static/riot/competitions/detail/_registration.tag
```diff
@@ -0,0 +1,95 @@
...
+                            .done(competition => {
+                                CODALAB.events.trigger('competition_loaded', competition)
+                            })
+                    } else {
+                        toastr.success('Your registration application is being processed!')
+                    }
+                    self.update()
+                })
+                .fail(response => {
+                    toastr.error('Something went wrong.')
```
ckcollab: better error?

src/static/riot/competitions/detail/participant_manager.tag
```diff
@@ -0,0 +1,108 @@
...
+                <th class="center aligned">Actions</th>
+            </tr>
+            </thead>
+            <tbody>
+            <tr each="{participants}">
+                <td>{username}</td>
+                <td>{email}</td>
+                <td>{_.startCase(status)}</td>
+                <td class="right aligned">
+                    <button class="mini ui red button icon" show="{status !== 'denied'}"
```
ckcollab: would prefer all html attributes line up, , so `show=` should be on its own line lining up with `class=` as well

src/static/riot/competitions/detail/participant_manager.tag
```diff
@@ -0,0 +1,108 @@
...
+                            data-tooltip="Approve"
+                            data-inverted=""
+                            data-position="bottom center">
+                            <i class="checkmark icon"></i>
+                    </button>
+                    <button class="mini ui blue button icon"
+                            data-tooltip="Send Message"
+                            data-inverted=""
+                            data-position="bottom center"
+                            onclick="">
```
ckcollab: pull this onclick? -- and other empty attributes around here?

src/static/riot/competitions/detail/participant_manager.tag
```diff
@@ -0,0 +1,108 @@
...
+            </tr>
+            </tbody>
+        </table>
+    </div>
+
+
+    <script>
+        let self = this
+
+
```
ckcollab: extra whitespace here and above

src/apps/competitions/emails.py
jimmykodes: added `all_organizers` as a property on the competition, which returns a list of the creator and all collabs. so just doing some list comprehension on that property instead.
`[user.email for user in comp.all_organizers]` not sure we necessarily need a property for an email list if we already have a list of users?

ckcollab: sounds good

src/static/riot/competitions/detail/participant_manager.tag
jimmykodes: pointed the `onlick` to a placeholder function for dealing w/ sending emails. (logic that doesn't exist yet'
the `data-inverted=''` has to remain there so that the tooltip will be dark instead of light. It's a weird Semantic thing

src/apps/api/views/competitions.py
jimmykodes: Not really sure what kind of permissions checks to throw in here?

ckcollab: Well right now anyone could access participant models for `GET`, `PUT`, `DELETE`

You could make queryset filter only participants of competitions they are owner/collaborator on? Not really a "permissions check" as it is a filter on what participants you have access to editing/viewing

src/static/riot/competitions/detail/_registration.tag
jimmykodes: Removing the href also removes the `hover` styles of the link. Could add those back manually, but i think `preventDefault` is the easier solution to the problem.

ckcollab:
:shipit: (knowing that we have to deal with terms + conditions stuff!)

PR #215, opened 2019-07-25
Dataset and task management Merge

jimmykodes:


ckcollab:
image

For this view, I think we can make this a bit prettier by "attaching" the selected tab to the tab menu:

```html
<div class="ui top attached tabular menu">
  <div class="active item">Tab</div>
</div>
<div class="ui bottom attached active tab segment">
  <p></p>
  <p></p>
</div>
```

When I change this it looks like...

image

ckcollab:
image

Should probably change this to one link `"Manage Datasets / Tasks"`

ckcollab:
image

Cool, I think this is a move in the right direction!

:shipit:

PR #216, opened 2019-07-25
made api_url for run_submission look for an env var

jimmykodes:


ckcollab:
...right after approving, I have an idea..! we may want a better variable name than `API_URL`, maybe `COMPETITIONS_API_URL` or something?

PR #217, opened 2019-07-31
V15 bundle handling

ckcollab:
# TODO

- [x] Add `metadata` to `input/res` in scoring step

```
# from v1.5 compute worker
prog_status = {
    'exitCode': exit_code,
    'ingestionExitCode': ingestion_program_exit_code,
    'elapsedTime': elapsedTime
}

...

with open(join(output_dir, 'metadata'), 'w') as f:
    f.write(yaml.dump(prog_status, default_flow_style=False))
```
- [x] Since we're saving scoring step `output/` now, add Trello task for that


src/apps/competitions/tasks.py
```diff
@@ -165,11 +166,12 @@ def run_submission(submission_pk, task_pk=None, is_scoring=False):
     run_arguments = {
         "submissions_api_url": os.environ.get('SUBMISSIONS_API_URL', "http://django/api"),
         "secret": submission.secret,
-        "docker_image": "python:3.7",
+        "docker_image": "codalab/codalab-legacy:py3",
```
jimmykodes: should we be hardcoding this?

src/apps/competitions/tasks.py
```diff
@@ -239,10 +241,14 @@ def get_data_key(obj, file_type, temp_directory, creator):
...
     if not field:
         return None
+    elif isinstance(field, datetime.date):
+        # turn the date into a datetime @ midnight that day
+        field = datetime.datetime.combine(datetime.date.today(), datetime.time())
     elif not isinstance(field, datetime.datetime):
         field = parser.parse(field)
     field = field.replace(tzinfo=now().tzinfo)
     return field
+# >>>>>>> develop
```
jimmykodes: pull

docker/compute_worker/compute_worker.py
```diff
@@ -395,7 +445,18 @@ def push_scores(self):
...
+    def push_output(self):
+        """Output is pushed at the end of both prediction and scoring steps."""
+        # V1.5 compatibility, write program statuses to metadata file
+        prog_status = {
+            'exitCode': self.program_exit_code,
+            'ingestionExitCode': self.ingestion_program_exit_code,
+            'elapsedTime': self.program_elapsed_time
+        }
+
+        with open(os.path.join(self.output_dir, 'metadata'), 'w') as f:
```
ckcollab: TODO for Eric: if this file already exists in output, from participant/organizer's program... raise some kind of `SubmissionException`!

docker/compute_worker/compute_worker.py
```diff
@@ -249,30 +274,49 @@ def _run_program_directory(self, program_dir, kind, can_be_output=False):
...
             # Try the new timeout feature
             '--stop-timeout={}'.format(self.execution_time_limit),
+
             # Don't allow subprocesses to raise privileges
             '--security-opt=no-new-privileges',
-            # Set the right volume
-            '-v', f'{program_dir}:/app',
+
+            # Set the volumes
+            # '-v', f'{program_dir}:/app',
```
jimmykodes: Pull this?

docker/compute_worker/compute_worker.py
```diff
@@ -249,30 +274,49 @@ def _run_program_directory(self, program_dir, kind, can_be_output=False):
...
+            # program here is either scoring program or submission, depends on if this ran during Prediction or Scoring
+            docker_cmd += ['-v', f'{os.path.join(self.root_dir, "program")}:/app/ingested_program']
+
+        if self.input_data:
+            # if kind == 'ingestion':
+            #     # TODO: Check with zhengying (already contacted him, waiting) on whether we need a hidden dir or not.
+            #     docker_cmd += ['-v', f'{os.path.join(self.root_dir, "input_data")}:/app/hidden']
+            # else:
+            #     # Unhidden
+            #     docker_cmd += ['-v', f'{os.path.join(self.root_dir, "input_data")}:/app/input_data']
```
jimmykodes: Pull this too?

src/apps/competitions/tasks.py
```diff
@@ -239,6 +251,9 @@ def get_data_key(obj, file_type, temp_directory, creator):
 def _get_datetime(field):
     if not field:
         return None
+    elif isinstance(field, datetime.date):
+        # turn the date into a datetime @ midnight that day
+        field = datetime.datetime.combine(datetime.date.today(), datetime.time())
```
jimmykodes: should this be `field = datetime.datetime.combine(field, datetime.time())` Otherwise I think this will always set field to be whatever day it was added, irrespective of the field value?

src/apps/competitions/tasks.py
```diff
@@ -494,7 +509,11 @@ def unpack_competition(competition_dataset_pk):
                     continue
                 phase1 = competition['phases'][i - 1]
                 phase2 = competition['phases'][i]
-                if phase2['start'] < phase1['end']:
+                if phase1['end'] is None:
+                    raise CompetitionUnpackingException(
+                        f'Phase: {phase1.get("name", phase2["index"])} must have an end time because it has a phase after it.'
```
jimmykodes: should be `phase1['index']` not `phase2`

ckcollab: great catch

src/apps/competitions/tasks.py
ckcollab: Lame, added tests so this shouldn't happen again -- don't want to spend a ton of time making them full proof but lmk if there are a couple other quick cases we should maybe add

PR #224, opened 2019-08-30
Comp editor updates

jimmykodes:
# Improvements include...

Old: 
image

New:
image

Old: 
image

New:
image

Old:
image

New:
image

Old:
image

New:
image


ckcollab:
It looks like the border on the top doesn't match the border on the text section?

image

ckcollab:
The column key isn't marked as required -- I can't click save or anything to see required fields?

image

ckcollab:
Maybe add example terms link/helper here?
image

src/static/riot/competitions/editor/_pages.tag
```diff
@@ -1,44 +1,57 @@
...
+                    <tr>
+                        <th colspan="2">Pages</th>
+                    </tr>
+                    </thead>
+                    <tbody>
+                    <tr each="{page, index in pages}">
+                        <td>{page.title}</td>
+                        <td class="right aligned">
+                            <a class="chevron">
+                                <sorting-chevrons data="{ pages }"
```
ckcollab: nice

src/static/riot/competitions/editor/_participation.tag
```diff
@@ -0,0 +1,53 @@
...
+    <script>
+        let self = this
+
+        self.data = {}
+
+        self.on('mount', () => {
+            self.markdown_editor = create_easyMDE(self.refs.terms)
+
+            $(':input', self.root).not('[type="file"]').not('button').not('[readonly]').each(function (i, field) {
+                this.addEventListener('keyup', self.form_updated)
```
ckcollab: do we need to clean up `addEventListeners`?

src/static/riot/competitions/editor/_phases.tag
```diff
@@ -11,34 +8,53 @@
...
+                    <tr>
+                        <th colspan="2">Phases</th>
+                    </tr>
+                    </thead>
+                    <tbody>
+                    <tr each="{phase, index in phases}">
+                        <td>{ phase.name }</td>
+                        <td class="right aligned">
+                            <a class="chevron">
+                                <sorting-chevrons data="{ phases }" index="{ index }"
```
ckcollab: should move `index=` to its own line

src/static/riot/competitions/editor/_phases.tag
```diff
@@ -84,6 +100,54 @@
...
+                        Advanced
+                        <i class="cogs icon"></i>
+                    </div>
+                    <div class="content">
+                        <div class="three fields">
+                            <div class="field">
+                                <label>
+                                    Execution Time Limit <span data-tooltip="In milliseconds, 600ms default if unset"
+                                                               data-inverted=""
+                                                               data-position="bottom center"><i
```
ckcollab: can we put `<i` and such on the same line, I think?

src/static/riot/competitions/editor/_phases.tag
```diff
@@ -84,6 +100,54 @@
...
+                                        class="help icon circle"></i></span>
+                                </label>
+                                <input type="number" name="max_submissions_per_day">
+                            </div>
+                            <div class="field">
+                                <label>
+                                    Max Submissions Per Person <span
+                                        data-tooltip="The maximum number of submissions any single user can make to the phase"
+                                        data-inverted=""
+                                        data-position="bottom center"><i
```
ckcollab: same here

src/static/riot/competitions/editor/_phases.tag
```diff
@@ -144,6 +203,8 @@
                     self.clear_form()
                 }
             })
+            $('.ui.checkbox', self.root).checkbox()
+            $('.ui.advanced-settings.accordion').accordion()
```
ckcollab: probably OK to not have `self.root` on this guy?

jimmykodes: Adding `self.root` will actually cause this to fail (just confirmed, to make sure I wasn't mis-remembering). I think it has something to do w/ the fact that this is not actually on root, but on one of root's children? I had thought that that didn't matter, since it was still _technically_ a part of root, but for what ever reason, adding `self.root` makes the accordion fail.

src/static/riot/competitions/editor/_participation.tag
jimmykodes: what to you mean "clean up"?

src/static/riot/competitions/editor/_phases.tag
ckcollab: Agh, probably not a big deal, but would love to not have a global init here -- would a `ref` make sense? `#id`?

src/static/riot/competitions/editor/_participation.tag
ckcollab: _Disclaimer: may not be worth spending much time on, since we're not a long running single page app, we have page refreshes._

By "clean up" I mean implement some kind of `removeEventListener` when we're unmounted/what-have-you. The current code _may_ add listeners without them being garbage collected over and over and over again -- but, I may be misunderstanding how things are setup. If we're hiding the element, maybe these listeners aren't destroyed properly? If the element is destroyed, the event listener should be destroyed with it?

https://stackoverflow.com/questions/446892/how-to-find-event-listeners-on-a-dom-node-when-debugging-or-from-the-javascript

Looks like we could call `getEventListeners(dom element)` to check to see if things aren't cleaned up? Again, probably not worth the time to get into

src/static/riot/competitions/editor/_phases.tag
jimmykodes: not sure why i didn't just do a ref in the first place... done now.

src/static/riot/competitions/editor/_participation.tag
jimmykodes: Just using the event listeners section of chrome console, looks like these event listeners go away when you navigate away from the page, but there are a lot of them on the comp editor page since all the tags are mounted together, and then just hidden in tabs. Could maybe clean this up so there aren't like 30 event listeners or whatever, but probably not necessary?

ckcollab: Yeah, probably not a huge worry now -- we'll keep an eye on it every time we manually do some `addEventListener` jazz

PR #226, opened 2019-09-18
updated login and sign up forms

jimmykodes:


src/templates/registration/login.html
```diff
@@ -2,45 +2,67 @@
...
-                               value="{{ form.username.value|default:""}}"
-                               maxlength="{{ form.username.field.max_length }}"
-                               required>
-                    </div>
+
+            <div class="field">
+                <a class="ui fluid basic button" href="{% url 'social:begin' 'chahub' %}">Log in with ChaHub</a>
+            </div>
+            <div class="ui horizontal divider" style="padding-bottom: 5px; padding-top: 5px;">
+                &nbsp;&nbsp;&nbsp;&nbsp; Or &nbsp;&nbsp;&nbsp;&nbsp;
```
ckcollab: ehhhhh should we not do inline styling and `&nbsp;` here?

src/templates/registration/signup.html
```diff
@@ -2,98 +2,106 @@
...
-                    {% endif %}
+            <div class="field">
+                <a class="ui fluid button" href="{% url 'social:begin' 'chahub' %}">Sign up with ChaHub</a>
+            </div>
+            <div class="field">
+                <a class="ui fluid basic button" href="{{ chahub_signup_url }}" target="_blank">Create a ChaHub
+                    Account</a>
+            </div>
+            <div class="ui horizontal divider" style="padding-bottom: 5px; padding-top: 5px;">
+                &nbsp;&nbsp;&nbsp;&nbsp; Or &nbsp;&nbsp;&nbsp;&nbsp;
```
ckcollab: same as above

src/templates/registration/login.html
jimmykodes: I couldn't get anything _but_ inline styling to work here, no idea why... I'll give it another go though. as far as `&nbsp;` goes, not sure how else to increase the gap around the 'Or'

PR #232, opened 2019-10-15
Bundle converter refactor

Tthomas63:


src/apps/competitions/unpacker/utils.py
```diff
@@ -0,0 +1,64 @@
...
+from dateutil import parser
+from django.core.files import File
+from django.utils import timezone
+
+from datasets.models import Data
+from competitions.unpacker.exceptions import CompetitionUnpackingException
+
+logger = logging.getLogger()
+
+def get_data_key(obj, file_type, temp_directory, creator, index):
```
ckcollab: should prefix this with `_` maybe? or remove the prefix `_`'s from below?

src/apps/competitions/unpacker/exceptions.py
```diff
@@ -0,0 +1,7 @@
+
+class CompetitionUnpackingException(Exception):
+    pass
+
+
+class CompetitionConversionException(CompetitionUnpackingException):
```
ckcollab: is this useful? we don't even raise it anywher eI don't see?

Tthomas63: Converter line 214, 257, 259. Should probably be moved with the refactor

PR #233, opened 2019-10-15
Queue management

Tthomas63:


src/apps/queues/models.py
```diff
@@ -0,0 +1,53 @@
...
+import uuid
+
+from apps.queues import rabbit
+
+
+class Queue(models.Model):
+    name = models.CharField(max_length=64)
+    vhost = models.UUIDField(unique=True)
+    is_public = models.BooleanField(default=False)
+    owner = models.ForeignKey(
```
jimmykodes: why is owner nullable? Shouldn't a Queue owner be required?

src/apps/queues/models.py
```diff
@@ -0,0 +1,53 @@
...
+    def __str__(self):
+        return self.name
+
+    @property
+    def broker_url(self):
+        from django.contrib.sites.models import Site
+        host = Site.objects.get_current().domain
+
+        if self.owner:
+            return "pyamqp://{}:{}@{}/{}".format(
```
jimmykodes: use f strings

src/apps/queues/rabbit.py
```diff
@@ -0,0 +1,79 @@
...
+
+from pyrabbit.api import Client
+
+from django.conf import settings
+from pyrabbit.http import HTTPError, NetworkError
+
+
+def _get_rabbit_connection():
+    """Helper giving us a rabbit connection from settings.BROKER_URL"""
+    host_with_port = "{}:{}/".format(settings.RABBITMQ_HOST, settings.RABBITMQ_MANAGEMENT_PORT)
```
jimmykodes: f strings

src/apps/queues/rabbit.py
```diff
@@ -0,0 +1,79 @@
...
+        # We found the user, no need to initialize
+        return False
+    except (HTTPError, NetworkError):
+        # User not found, needs initialization
+        return True
+
+
+def initialize_user(user):
+    """Check whether user has a rabbitmq account already, creates it if not."""
+    print("Making new rabbitmq user for {}".format(user))
```
jimmykodes: logger? or pull? also, f strings if keeping

src/templates/queues/form.html
```diff
@@ -0,0 +1,30 @@
...
+
+{% block head_title %}Queue form{% endblock %}
+{% block page_title %}<h2>Queue form</h2>{% endblock page_title %}
+
+{% block content %}
+    <div class="ui container message">
+        <form class="ui form" action="" method="post">
+            {% csrf_token %}
+            <table class="ui table table-bordered">
+                {{ form.as_table }}
```
jimmykodes: why are we doing this with forms? Shouldn't we be doing this like everything else in v2? with the API?

src/settings/base.py
```diff
@@ -196,6 +199,8 @@
 # Celery
 # ============================================================================
 CELERY_BROKER_URL = os.environ.get("CLOUDAMQP_URL") or os.environ.get('BROKER_URL')
+# Rabbit/Pyrabbit reference BROKER_URL internally
+BROKER_URL = CELERY_BROKER_URL
```
jimmykodes: Is there no better workaround than this? If we _have_ to do it this way, we probably need to do this assignment after the `if not CELERY_BROKER_URL:` stuff, right? so this gets set to the correct value?

src/settings/base.py
```diff
@@ -176,7 +177,9 @@
 # =============================================================================
 # SSL
 # =============================================================================
-if os.environ.get('USE_SSL'):
+USE_SSL = os.environ.get('USE_SSL')
```
jimmykodes: are we referencing `USE_SSL` anywhere else? does this need to be a variable?

src/apps/queues/views.py
```diff
@@ -0,0 +1,76 @@
+# from braces.views import LoginRequiredMixin
```
jimmykodes: pull

src/apps/queues/views.py
```diff
@@ -0,0 +1,76 @@
...
+from django.contrib.auth.mixins import LoginRequiredMixin
+from django.core.exceptions import PermissionDenied, ValidationError
+# from django.core.urlresolvers import reverse
```
jimmykodes: pull

src/apps/queues/rabbit.py
```diff
@@ -0,0 +1,79 @@
...
+from pyrabbit.api import Client
+
+from django.conf import settings
+from pyrabbit.http import HTTPError, NetworkError
+
+
+def _get_rabbit_connection():
+    """Helper giving us a rabbit connection from settings.BROKER_URL"""
+    host_with_port = "{}:{}/".format(settings.RABBITMQ_HOST, settings.RABBITMQ_MANAGEMENT_PORT)
+    # TODO: Scheme was deprecated? Verify this
```
jimmykodes: this TODO?

src/apps/competitions/models.py
```diff
@@ -28,6 +29,8 @@ class Competition(ChaHubSaveMixin, models.Model):
     description = models.TextField(null=True, blank=True)
     docker_image = models.CharField(max_length=128, default="codalab/codalab-legacy:py3")
 
+    queue = models.ForeignKey(Queue, on_delete=models.SET_NULL, null=True, blank=True, related_name='competitions')
```
jimmykodes: How does this value get changed? I don't see anything additions to the comp editor?

src/apps/queues/forms.py
```diff
@@ -0,0 +1,27 @@
...
+
+        # Remove extra help text that obfuscates our organizer message
+        remove_message = 'Hold down "Control", or "Command" on a Mac, to select more than one.'
+        self.fields['organizers'].help_text = self.fields['organizers'].help_text.replace(remove_message, '')
+
+    def clean(self):
+        # If we're creating this, make sure we don't have > the limited # of queues
+        if not self.instance.pk:
+            queue_count = Queue.objects.filter(owner=self.user).count()
+            if queue_count >= self.user.rabbitmq_queue_limit:
```
jimmykodes:
```python
if self.user.queues.count() >= self.user.rabbitmq_queue_limit:
# might need all() too...
self.user.queues.all().count()
```
Pretty sure you don't need the `all()`, but you might.

src/apps/queues/forms.py
```diff
@@ -0,0 +1,27 @@
...
+                raise ValidationError("Cannot create more than {} queues".format(self.user.rabbitmq_queue_limit))
```
jimmykodes: f string

src/apps/queues/forms.py
```diff
@@ -0,0 +1,27 @@
...
+
+class QueueForm(ModelForm):
+
+    class Meta:
+        model = Queue
+        fields = ('name', 'is_public', 'organizers')
+
+    def __init__(self, *args, **kwargs):
+        self.user = kwargs.pop('user')
+        super(QueueForm, self).__init__(*args, **kwargs)
```
jimmykodes: python 3 `super` doesn't require these args

src/apps/queues/forms.py
```diff
@@ -0,0 +1,27 @@
...
+        remove_message = 'Hold down "Control", or "Command" on a Mac, to select more than one.'
+        self.fields['organizers'].help_text = self.fields['organizers'].help_text.replace(remove_message, '')
+
+    def clean(self):
+        # If we're creating this, make sure we don't have > the limited # of queues
+        if not self.instance.pk:
+            queue_count = Queue.objects.filter(owner=self.user).count()
+            if queue_count >= self.user.rabbitmq_queue_limit:
+                raise ValidationError("Cannot create more than {} queues".format(self.user.rabbitmq_queue_limit))
+        return super(QueueForm, self).clean()
```
jimmykodes: same `super` args.

src/apps/queues/models.py
```diff
@@ -0,0 +1,53 @@
...
+        on_delete=models.DO_NOTHING,
+        related_name='queues',
+        blank=True,
+        null=True
+    )
+    organizers = models.ManyToManyField(
+        settings.AUTH_USER_MODEL,
+        related_name='organized_queues',
+        blank=True,
+        help_text="(Organizers allowed to view this queue when they assign their competition to a queue)"
```
jimmykodes: do we need this? seems superfluous, we don't add help text anywhere else

src/apps/queues/models.py
```diff
@@ -0,0 +1,53 @@
...
+    )
+
+    def __str__(self):
+        return self.name
+
+    @property
+    def broker_url(self):
+        from django.contrib.sites.models import Site
```
jimmykodes: Why import down here?

src/apps/queues/views.py
```diff
@@ -0,0 +1,76 @@
...
+from django.http import HttpResponseRedirect
+from django.views.generic import ListView, CreateView, UpdateView, DeleteView
+from pyrabbit.http import HTTPError
+
+from .forms import QueueForm
+from .models import Queue
+from . import rabbit
+
+
+class QueueFormMixin(object):
```
jimmykodes: don't need `object`

src/apps/queues/views.py
```diff
@@ -0,0 +1,76 @@
...
+    def get_success_url(self):
```
jimmykodes: why do we need this function? why can't we just use `reverse` each time?

src/apps/queues/views.py
```diff
@@ -0,0 +1,76 @@
...
+        return reverse('queues:list')
+
+    def get_form_kwargs(self, *args, **kwargs):
+        kwargs = super(QueueFormMixin, self).get_form_kwargs(*args, **kwargs)
```
jimmykodes: don't need `super` args

ckcollab:
> Before I keep going into more comments of forms stuff, why are we using forms for this? instead of the API?

(summary of our convo) I think it's OK to do this the same style we did in v1.5, and leave it as a nice ["beginner" issue](https://github.com/codalab/competitions-v2/issues/234) to re-do DRF style

src/apps/queues/models.py
ckcollab: If someone makes a queue, attaches it to a competition, then deletes their account, the idea is that this queue will stay alive and competitions using it won't be harmed

src/settings/base.py
ckcollab: I don't think pyrabbit even needs this? No where in the code does it reference this, unless I am mistaken?

https://github.com/bkjones/pyrabbit/search?q=broker_url&unscoped_q=broker_url

src/apps/queues/views.py
```diff
@@ -0,0 +1,76 @@
...
+class QueueCreateView(LoginRequiredMixin, QueueFormMixin, CreateView):
+    model = Queue
+    template_name = 'queues/form.html'
+    form_class = QueueForm
+
+    def form_valid(self, form):
+        try:
+            queue = form.save(commit=False)
+            queue.owner = self.request.user
+            if rabbit.check_user_needs_initialization(self.request.user):
```
jimmykodes: What do you think about doing this logic _inside_ `create_queue` instead? Then, if we ever call `create_queue` from somewhere else, we always know the user we pass to it will be initialized? Would clean up this code quite a bit, too.

src/apps/queues/views.py
```diff
@@ -0,0 +1,76 @@
...
+                rabbit.initialize_user(self.request.user)
+            queue.vhost = rabbit.create_queue(self.request.user)
+
+            # Only save queue if things were successful
+            queue.save()
+            form.save_m2m()
```
jimmykodes: maybe add small comment here that this is for saving the ManyToMany relationships from the form? since we don't use forms super often, a little extra commenting will make maintaining this nicer?

src/apps/queues/views.py
```diff
@@ -0,0 +1,76 @@
...
+            return HttpResponseRedirect(self.get_success_url())
+        except HTTPError:
+            errors = form._errors.setdefault("__all__", ErrorList())
```
jimmykodes: Also, maybe comment here about `necessary weirdness to inject additional non-field errors` or whatever. Help those of us who aren't forms-fluent down the road?

src/apps/queues/views.py
```diff
@@ -0,0 +1,76 @@
...
+class QueueListView(LoginRequiredMixin, ListView):
+    model = Queue
+    template_name = 'queues/list.html'
+
+    def get_queryset(self):
+        qs = Queue.objects.filter(owner=self.request.user) | self.request.user.organized_queues.all()
+        return qs
+
+
+class QueueCreateView(LoginRequiredMixin, QueueFormMixin, CreateView):
```
jimmykodes: Do these all need to be separate views? Can we combine Create and Update (and maybe delete? though that seems unlikely) with a FormView?

src/templates/queues/base_queues.html
```diff
@@ -0,0 +1,14 @@
+{% extends "base.html" %}
+
+{% block extra_headers %}
+    <link rel="stylesheet" href="{{ STATIC_URL }}forums/forums.css">
```
jimmykodes: Should this be done with `{% load static %} ... {% static url ... %}` instead? Also, is forums.css the right one to load in?

src/templates/queues/delete.html
```diff
@@ -0,0 +1,17 @@
+{% extends "queues/base_queues.html" %}
+
+{% block head_title %}Queue delete{% endblock %}
+{% block page_title %}<h2>Queue delete</h2>{% endblock page_title %}
```
jimmykodes: I don't think this block even exists?

src/templates/queues/delete.html
```diff
@@ -0,0 +1,17 @@
...
+
+{% block content %}
+    <div class="ui container message">
+        <form class="ui form" action="" method="post" style="text-align: center;">
+            {% csrf_token %}
+
+            <h3>Are you sure you want to delete this queue?</h3>
+
+            <input type="submit" value="Confirm delete" class="ui red button">
+            <a href="{% url "queues:list" %}" class="ui button" style="margin-right: 10px;">Cancel</a>
```
jimmykodes: inline style.

src/templates/queues/delete.html
```diff
@@ -0,0 +1,17 @@
...
+{% block page_title %}<h2>Queue delete</h2>{% endblock page_title %}
+
+{% block content %}
+    <div class="ui container message">
+        <form class="ui form" action="" method="post" style="text-align: center;">
+            {% csrf_token %}
+
+            <h3>Are you sure you want to delete this queue?</h3>
+
+            <input type="submit" value="Confirm delete" class="ui red button">
```
jimmykodes: `<input>` not `<button>` ?

src/templates/queues/delete.html
```diff
@@ -0,0 +1,17 @@
+{% extends "queues/base_queues.html" %}
+
+{% block head_title %}Queue delete{% endblock %}
+{% block page_title %}<h2>Queue delete</h2>{% endblock page_title %}
+
+{% block content %}
+    <div class="ui container message">
+        <form class="ui form" action="" method="post" style="text-align: center;">
```
jimmykodes: inline style

src/templates/queues/form.html
```diff
@@ -0,0 +1,30 @@
...
+            </table>
+
+            <input type="submit" value="Save" class="ui blue button" style="float: right;">
```
jimmykodes: inline styling. And, button?

src/templates/queues/form.html
```diff
@@ -0,0 +1,30 @@
...
+
+    <style>
+        .select2-container {
+            width: 100%;
+        }
+    </style>
+
+    <script type="text/javascript">
+        $('document').ready(function () {
+            $("#id_organizers").select2()
```
jimmykodes: should probably rework this to use semantics multiselect? instead of loading in a new library for this. also, where are we loading this library in?

src/templates/queues/list.html
```diff
@@ -0,0 +1,46 @@
+{% extends "queues/base_queues.html" %}
+
+{% block head_title %}Queues list{% endblock %}
+{% block page_title %}<h2>Queues list</h2>{% endblock page_title %}
+
+{% block content %}
+    <div class="ui message container">
+        <a href="https://github.com/codalab/codalab-competitions/wiki/Dev-Worker-Queues"><i class="ui question icon"></i> Read more about Worker Queues here.</a>
+
+        <a href="{% url "queues:create" %}" class="ui blue button" style="float: right; margin: 10px 0 10px 10px;">
```
jimmykodes: inline styling

src/templates/queues/list.html
```diff
@@ -0,0 +1,46 @@
...
+            Add Queue
+        </a>
+
+        <table class="ui celled striped table">
+            <thead>
+                <tr style="font-weight: bold;">
```
jimmykodes: inline styling

src/templates/queues/list.html
```diff
@@ -0,0 +1,46 @@
...
+                        {% if queue in request.user.queues.all %}
+                            <a href="{% url "queues:update" pk=queue.pk %}" class="ui button">Edit</a>
+                            <a href="{% url "queues:delete" pk=queue.pk %}" class="ui red button">Delete</a>
+                        {% endif %}
+                    </td>
+                </tr>
+            {% empty %}
+                <tr>
+                    <td colspan="4">
+                        <i style="display: block; padding: 20px; text-align: center;">No queues created, yet!</i>
```
jimmykodes: inline styling. Also, should use `<em>` not `<i>`

src/templates/queues/list.html
```diff
@@ -0,0 +1,46 @@
...
+                    <td width="100px;">Is public?</td>
+                    <td width="150px;">Actions</td>
+                </tr>
+            </thead>
+            <tbody>
+            {% for queue in object_list %}
+                <tr>
+                    <td>{{ queue.name }}</td>
+                    <td>{{ queue.broker_url }}</td>
+                    <td>{% if queue.is_public %}<i class="glyphicon glyphicon-ok"></i>{% endif %}</td>
```
jimmykodes: use semantic icons, not glyphicons

src/templates/queues/form.html
Tthomas63: Do we even want a multiselect 2 here or is it fine to just pull?

src/apps/queues/rabbit.py
Tthomas63: Tested this and I'm not 100% but I think I can pull it. Signature of the class doesn't seem to take it, and it works with/without...

src/settings/base.py
ckcollab: why was this marked resolved?

Tthomas63: I pulled this. I could've sworn PyRabbit was referencing it, but it does not appear to be.

src/apps/queues/models.py
jimmykodes: I vote pull it, since we don't use it anywhere else. (I didn't even know it was an option...)

src/apps/queues/views.py
jimmykodes: It looks like `get_success_url` looks for either `self.success_url` or `self.object.get_absolute_url()`
so instead of overwriting this function, we could just set the variable? like:
```python
class QueueFormMixin:
    success_url = reverse('queues:list')
```
What do you think about that instead? or do we wanna override the function? I could kinda go either way on it, It just feels weird to me that we are inheriting from `object` and overwriting the functions of other classes... Maybe I'm just biased against mixins....

src/apps/queues/views.py
jimmykodes: Hmm. I'm thinking we take just a quick look into using FormView, and if it looks like it'll be all-encompassing, than switch. But I suppose there is some benefit to the explicitness of having these all broken out. I think I'm just used to looking at ViewSets where these are abstracted away from us, so it seems weird to have it explicit.

Tthomas63:
image
image

src/apps/competitions/models.py
Tthomas63: Added to comp editor

src/static/riot/competitions/editor/_competition_details.tag
```diff
@@ -94,10 +157,16 @@
...
             self.refs.title.value = competition.title
             self.markdown_editor.value(competition.description || '')
+            self.refs.queue.value = _.get(competition, 'queue.id', null)
+            $('.queue-search').dropdown('set text', _.get(competition, 'queue.name', null))
+            $('.queue-search').dropdown('set value', _.get(competition, 'queue.id', null))
 
             // Value comes like c:/fakepath/file_name.txt -- cut out everything but file_name.txt
-            self.uploaded_logo_name = competition.logo.replace(/\\/g, '/').replace(/.*\//, '')
-            self.uploaded_logo = competition.logo
+            // TODO: Added this because my form was not receiving a logo from compeititon.
```
jimmykodes: Did the competition have a logo? is this going to mask a different error?

src/static/riot/queues/list.tag
```diff
@@ -0,0 +1,282 @@
...
+                        name: 'username',
+                        value: 'id',
+                    },
+                    cache: false,
+                    maxResults: 5,
+                })
+            ;
+        })
+
+        self.switch_to_form = () => {
```
jimmykodes: Is this still necessary?

src/static/riot/queues/list.tag
```diff
@@ -0,0 +1,282 @@
...
+    <script>
+        var self = this
+        self.queues = []
+        self.selected_queue = {}
+        self.page = 1
+
+        self.one("mount", function () {
+            self.update_queues()
+            $(".ui.checkbox", self.root).checkbox()
+            $('.collab-search')
```
jimmykodes: `ref` instead

src/static/riot/queues/list.tag
```diff
@@ -0,0 +1,282 @@
...
+                .dropdown({
```
jimmykodes: on the same line as jquery selector

src/static/riot/queues/list.tag
```diff
@@ -0,0 +1,282 @@
...
+                    preserveHTML: false,
+                    fields: {
+                        remoteValues: 'results',
+                        name: 'username',
+                        value: 'id',
+                    },
+                    cache: false,
+                    maxResults: 5,
+                })
+            ;
```
jimmykodes: don't need this `;`

src/static/riot/queues/list.tag
```diff
@@ -0,0 +1,282 @@
...
+        self.close_modal = () => {
+            $(self.refs.modal).modal('hide')
+            self.clear_form()
+        }
+
+        self.clear_form = () => {
+            $('.collab-search').dropdown('clear')
+            self.refs.queue_name.value = ''
+            self.selected_queue = {}
+            self.refs.queue_public.checked = false
```
jimmykodes: does this work? you don't have to do `$(self.refs.).prop('checked', false)`?

src/static/riot/queues/list.tag
```diff
@@ -0,0 +1,282 @@
...
+        self.handle_queue = function () {
+            let data = {
+                name: self.refs.queue_name.value,
+                is_public: self.refs.queue_public.checked,
+                organizers: _.compact($('.collab-search').dropdown('get value').split(','))
+            }
+            let endpoint = !_.isEmpty(self.selected_queue)
+                ? CODALAB.api.update_queue
+                : CODALAB.api.create_queue
+            endpoint(data, _.get(self, 'selected_queue.id'))
```
jimmykodes: this will make more sense as `_.get(self.selected_queue, 'id')`

src/static/riot/queues/list.tag
```diff
@@ -0,0 +1,282 @@
...
+                .done(function (response) {
+                    toastr.success("Succesfully updated/made queue!")
```
jimmykodes: This success message wording should either be changed to be agnostic of whether update or create was called. Or we should determine which was done and display the correct message.

src/static/riot/queues/list.tag
```diff
@@ -0,0 +1,282 @@
...
+                    self.close_modal()
+                    self.update_queues()
+                })
+                .fail(function (response) {
+                    toastr.error("Could not update/create queue!")
```
jimmykodes: Same with this error message

src/static/riot/queues/list.tag
```diff
@@ -0,0 +1,282 @@
...
+            } else {
+                alert("No valid page to go to!")
+            }
+        }
+        self.previous_page = function () {
+            if (!!self.pagination.previous) {
+                self.page -= 1
+                self.filter({page: self.page})
+            } else {
+                alert("No valid page to go to!")
```
jimmykodes: Probably don't raise alerts?

src/static/riot/queues/list.tag
```diff
@@ -0,0 +1,282 @@
...
+            self.page = filters.page
+            self.update_queues(filters)
+        }
+
+        self.next_page = function () {
+            if (!!self.pagination.next) {
+                self.page += 1
+                self.filter({page: self.page})
+            } else {
+                alert("No valid page to go to!")
```
jimmykodes: Probably don't raise alerts? Buttons should be hidden anyway, right?

src/static/riot/queues/list.tag
```diff
@@ -0,0 +1,282 @@
...
+            if (!!self.pagination.previous) {
+                self.page -= 1
+                self.filter({page: self.page})
+            } else {
+                alert("No valid page to go to!")
+            }
+        }
+
+        self.delete_queue = function (queue) {
+            if (confirm("Are you sure you want to delete '" + queue.name + "'?")) {
```
jimmykodes: Use template string

src/static/riot/queues/list.tag
```diff
@@ -0,0 +1,282 @@
...
+                </span>
+                <span class="popup" data-tooltip="Delete Queue">
+                    <i class="red icon trash alternate outline popup-button" if="{ queue.is_owner && !!queue.broker_url }"
+                       onclick="{ delete_queue.bind(this, queue) }"></i>
+                </span>
+            </td>
+        </tr>
+
+        <tr if="{queues.length === 0}">
+            <td class="center aligned" colspan="4">
```
jimmykodes: `colspan=5` ?

src/static/riot/queues/list.tag
```diff
@@ -0,0 +1,282 @@
...
+                <em>No Queues Yet!</em>
+            </td>
+        </tr>
+        </tbody>
+        <tfoot>
+        <!-------------------------------------
+                  Pagination
+        ------------------------------------->
+        <tr if="{queues.length > 0 && _.get(pagination, 'next')}">
+            <th colspan="6">
```
jimmykodes: `colspan=5` ?

src/static/riot/queues/list.tag
```diff
@@ -0,0 +1,282 @@
...
+        <tbody>
+        <tr each="{ queue in queues }" class="queue-row">
+            <td>{ queue.name }</td>
+            <td>{ queue.owner }</td>
+            <td>{ timeSince(Date.parse(queue.created_when)) } ago</td>
+            <td class="center aligned">
+                <i class="checkmark box icon green" if="{ queue.is_public }"></i>
+            </td>
+            <td class="right aligned">
+                <span class="popup" data-tooltip="{queue.broker_url}" data-position="left center">
```
jimmykodes: do you still need `class="popup"` on all of these?

src/static/riot/queues/list.tag
```diff
@@ -0,0 +1,282 @@
...
+        </tfoot>
+    </table>
+
+    <div class="ui modal" ref="modal">
+        <div class="header">
+            Queue Form
+        </div>
+        <div class="content">
+            <form class="ui form" ref="form">
+                <div class="ui active tab" data-tab="details">
```
jimmykodes: what is this for? do we have a menu here?

src/static/riot/queues/list.tag
```diff
@@ -0,0 +1,282 @@
...
+                CODALAB.api.delete_queue(queue.id)
+                    .done(function () {
+                        self.update_queues()
+                        toastr.success("Queue deleted successfully!")
+                    })
+                    .fail(function () {
+                        toastr.error("Could not delete queue!")
+                    })
+            }
+            event.stopPropagation()
```
jimmykodes: what event are we stopping propagation on?

src/static/riot/queues/list.tag
```diff
@@ -0,0 +1,282 @@
...
+        }
+
+        self.copy_queue_url = function (queue) {
+            navigator.clipboard.writeText(queue.broker_url).then(function () {
+                /* clipboard successfully set */
```
jimmykodes: change these to be line comments, not block?

src/templates/queues/list.html
```diff
@@ -0,0 +1,7 @@
+{% extends "queues/base_queues.html" %}
```
jimmykodes: this should just extend base. Don't think we need the intermediary template

src/apps/api/serializers/queues.py
```diff
@@ -0,0 +1,75 @@
...
+        ]
+
+
+class QueueSerializer(serializers.ModelSerializer):
+    is_owner = serializers.SerializerMethodField()
+    organizers = serializers.PrimaryKeyRelatedField(queryset=User.objects.all(), many=True, required=False)
+
+    class Meta:
+        model = Queue
+        fields = [
```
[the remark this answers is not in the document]
Tthomas63: DRF themselves actually switched to lists on their repo documentation this year, but haven't updated their other docs. This issue and the related PR discussions seem to indicate the reason for this was being consistent with Django, but Django themselves updated to using lists instead of tuples. If you think these should be tuples I'll change them.

https://github.com/encode/django-rest-framework/issues/6604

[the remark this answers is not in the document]
ckcollab: i think tuples make more sense because it should be immutable?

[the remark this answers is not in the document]
jimmykodes: That seems to be the heart of the argument in Tyler's referenced DRF issue, they aren't necessarily immutable.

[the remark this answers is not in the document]
jimmykodes: That said, we pretty much use tuples everywhere else in the code, and we aren't doing any mutations, sooo? for consistency, tuples, but for what is being argued as "best practice" list... kinda a coin toss?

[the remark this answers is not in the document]
Tthomas63: I'll just change to tuples

src/apps/competitions/models.py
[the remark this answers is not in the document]
Tthomas63: I have had issues with migrations before when referencing ForeignKeys through strings, but I will update it to reference it this way.

The exact use case was on Chagrade when I later updated the ForeignKey, the new migrations weren't able to find the reference for some reason... (I can probably still find the conversation but it took a minute to figure out what was going wrong and why)

src/apps/queues/urls.py
```diff
@@ -0,0 +1,9 @@
+from django.urls import path
+from queues import views
```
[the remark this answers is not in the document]
Tthomas63: I've always heard to use explicit imports vs local imports in Python. I'm not sure what our best practice is on this, but I will change it if you think we should be just doing a local import here.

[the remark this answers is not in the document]
ckcollab: relative imports are fine afaik, def prefer them in these cases

src/apps/queues/models.py
```diff
@@ -0,0 +1,47 @@
...
+from queues import rabbit
+
+
+class Queue(models.Model):
+    name = models.CharField(max_length=64)
+    vhost = models.UUIDField(unique=True)
+    is_public = models.BooleanField(default=False)
+    owner = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.DO_NOTHING,
```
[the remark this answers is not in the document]
Tthomas63: I think Eric and I originally thought that we should use `models.DO_NOTHING` because it should still contain a PK reference to the user who made it? (v1.5 was doing something weird in regards to this that we fixed recently, or have a PR open for it.) But setting `null` also makes sense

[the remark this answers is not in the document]
jimmykodes: I may be wrong, but _pretty_ sure that postgres won't let you delete a user who still has queue objects pointed to them because of integrity enforcement, so this should be something else?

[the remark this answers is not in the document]
Tthomas63: I'll change it to `models.SET_NULL`.

I believe though for that integrity enforcement, we'd have to set `on_delete=models.PROTECT`.
v1.5 issue/PR:
https://github.com/codalab/codalab-competitions/issues/2606
https://github.com/codalab/codalab-competitions/pull/2693

src/static/riot/competitions/editor/_competition_details.tag
Tthomas63: The competition did have a logo.

I saw a comment I believe in regards to this:
```
            // Require title, logo is optional IF we are editing -- will just keep the old one if
            // a new one is not provided
```

But it seems like this is something to look into

src/static/riot/queues/list.tag
Tthomas63: I believe this does work

src/apps/queues/models.py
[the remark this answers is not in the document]
jimmykodes: Django docs:
image

Proven by hand:
image

src/static/riot/queues/list.tag
Tthomas63: No I missed this when I carried over the HTML for the modal from tasks. Pulling it

src/static/riot/competitions/editor/_competition_details.tag
```diff
@@ -57,6 +70,31 @@
...
+                    //cache: false,
+                    maxResults: 5,
+                    onChange: (value, title) => {
+                        self.refs.queue.value = value
+                        self.form_updated()
+                    }
+                })
+            ;
+
+            self.get_available_queues()
```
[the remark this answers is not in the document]
Tthomas63: Wanted to load some initial options, but you're right in this doesn't make sense with how search/select works. Pulled

src/static/riot/competitions/editor/_competition_details.tag
```diff
@@ -86,6 +125,30 @@
...
+                search: $(self.refs.queue_search).val(),
+                page: 1,
+            })
+            self.page = filters.page
+            self.get_available_queues(filters)
+        }
+
+        self.get_available_queues = function(filters) {
+            filters = filters || {}
+            filters.public = true
```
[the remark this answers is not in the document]
Tthomas63: Pulled this, but the reasoning was I had not added a checkbox or anything to display public here on the competition details tag. I thought it made the most sense to default to showing public, that way all available options to the user were there.

src/apps/api/views/queues.py
```diff
@@ -0,0 +1,62 @@
...
+        show_public = self.request.query_params.get('public')
+        qs = self.queryset.prefetch_related('organizers')
+        if show_public:
+            qs = qs.filter(Q(is_public=True) | Q(owner=self.request.user)) | self.request.user.organized_queues.all()
+        else:
+            qs = qs.filter(owner=self.request.user) | self.request.user.organized_queues.all()
+        # Distinct so that we don't get duplicates
+        return qs.order_by('-created_when').distinct()
+
+    def get_serializer_class(self):
```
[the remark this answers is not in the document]
Tthomas63: This makes sense to me, updated the names.

src/apps/api/views/queues.py
```diff
@@ -0,0 +1,62 @@
...
+        if self.request.method == 'GET':
+            return serializers.QueueDetailSerializer
+        else:
+            return serializers.QueueSerializer
+
+    def get_serializer_context(self):
```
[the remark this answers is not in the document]
Tthomas63: Changed this as well, talked to you in person about some weirdness where nested serializers do not get context passed through if it's not a WritableNested

src/apps/competitions/tasks.py
```diff
@@ -149,8 +151,20 @@ def _send_submission(submission, task, is_scoring, run_args):
...
+    if submission.phase.competition.queue:
+        submission.queue_name = submission.phase.competition.queue.name or ''
+        submission.save()
+
+        # Send to special queue? Using `celery_app` var name here since we'd be overriding the imported `app` variable above
+        celery_app = app_or_default()
+        with celery_app.connection() as new_connection:
+            new_connection.virtual_host = str(submission.phase.competition.queue.vhost)
+            task = celery_app.send_task('compute_worker_run', args=(run_args,), queue='compute-worker',
+                                 soft_time_limit=time_limit, connection=new_connection)
```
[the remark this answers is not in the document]
Tthomas63: Broke it out onto newlines like so:
```
    if submission.phase.competition.queue:
        submission.queue_name = submission.phase.competition.queue.name or ''
        submission.save()

        # Send to special queue? Using `celery_app` var name here since we'd be overriding the imported `app` 
        # variable above
        celery_app = app_or_default()
        with celery_app.connection() as new_connection:
            new_connection.virtual_host = str(submission.phase.competition.queue.vhost)
            task = celery_app.send_task(
                'compute_worker_run',
                args=(run_args,),
                queue='compute-worker',
                soft_time_limit=time_limit,
                connection=new_connection
            )
    else:
        task = app.send_task(
            'compute_worker_run',
            args=(run_args,),
            queue='compute-worker',
            soft_time_limit=time_limit
        )
    submission.task_id = task.id
    submission.status = Submission.SUBMITTED
    submission.save()
```

src/apps/queues/models.py
```diff
@@ -0,0 +1,47 @@
...
+        related_name='queues',
+        blank=True,
+        null=True
+    )
+    organizers = models.ManyToManyField(
+        settings.AUTH_USER_MODEL,
+        related_name='organized_queues',
+        blank=True,
+    )
+    created_when = models.DateTimeField(auto_now_add=True)
```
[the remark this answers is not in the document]
Tthomas63: Updated to:
```
created_when = models.DateTimeField(default=timezone.now)
```

Makes sense for test cases and such.

src/apps/queues/models.py
```diff
@@ -0,0 +1,47 @@
...
+        if self.owner:
+            return f"pyamqp://{self.owner.rabbitmq_username}:{self.owner.rabbitmq_password}@{host}/{self.vhost}"
+
+    def delete(self, using=None):
+        try:
+            rabbit.delete_vhost(str(self.vhost))
+        except HTTPError:
+            # Vhost not found or something
+            pass
+        return super(Queue, self).delete(using)
```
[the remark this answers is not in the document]
Tthomas63: Missed this one, it has been pulled.

src/apps/queues/rabbit.py
```diff
@@ -0,0 +1,82 @@
...
+    user.save()
+
+
+def create_queue(user):
+    """Create a new queue with a random name and give full permissions to the owner AND our base account"""
+    if check_user_needs_initialization(user):
+        initialize_user(user)
+
+    vhost = str(uuid.uuid4())
+    conn = _get_rabbit_connection()
```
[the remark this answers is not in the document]
Tthomas63: It looked like the only calls to these helper functions were from `create_queue` itself. So I added another argument for a connection that can be passed from the calls from `create_queue`. So I believe now when `create_queue` is called, it should only call `_get_rabbit_connection` once and pass through the `connection` object to the helper functions. If we ever need to call the helper functions separately we can just manually call `_get_rabbit_connection` before hand I suppose?

```
def check_user_needs_initialization(user, connection):
    # conn = _get_rabbit_connection()
    try:
        connection.get_user_permissions(user.rabbitmq_username)
        # We found the user, no need to initialize
        return False
    except (HTTPError, NetworkError):
        # User not found, needs initialization
        return True


def initialize_user(user, connection):
    """Check whether user has a rabbitmq account already, creates it if not."""
    logger.info(f"Making new rabbitmq user for {user}")
    user.rabbitmq_username = str(uuid.uuid4())
    user.rabbitmq_password = str(uuid.uuid4())

    # conn = _get_rabbit_connection()
    connection.create_user(str(user.rabbitmq_username), str(user.rabbitmq_password))

    # Give user permissions to send submission updates
    connection.set_vhost_permissions(
        '/',
        str(user.rabbitmq_username),
        '.*',
        '.*submission-updates.*',
        '.*submission-updates.*'
    )

    # Was successful, save now
    user.save()


def create_queue(user):
    """Create a new queue with a random name and give full permissions to the owner AND our base account"""
    conn = _get_rabbit_connection()
    if check_user_needs_initialization(user, conn):
        initialize_user(user, conn)

    vhost = str(uuid.uuid4())
    conn.create_vhost(vhost)

    # Set permissions for our end user
    conn.set_vhost_permissions(
        vhost,
        user.rabbitmq_username,
        '.*',
        '.*',
        '.*'
    )

    # Set permissions for ourselves
    conn.set_vhost_permissions(
        vhost,
        settings.RABBITMQ_DEFAULT_USER,
        '.*',
        '.*',
        '.*'
    )

    return vhost
```

src/apps/queues/urls.py
```diff
@@ -0,0 +1,9 @@
...
+
+app_name = "queues"
+
+
+urlpatterns = [
+    path('', views.QueueManagementView.as_view(), name='list'),
```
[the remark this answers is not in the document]
Tthomas63: This does make more sense after the refactor. Updated the template names, and tags.

src/apps/queues/views.py
```diff
@@ -0,0 +1,7 @@
+from django.contrib.auth.mixins import LoginRequiredMixin
+from django.views.generic import TemplateView
+
+
+class QueueManagementView(LoginRequiredMixin, TemplateView):
+    template_name = 'queues/list.html'
```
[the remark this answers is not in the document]
Tthomas63: Updated to: `management.html`

src/static/js/ours/client.js
```diff
@@ -237,6 +237,25 @@ CODALAB.api = {
         return CODALAB.api.request('POST', `${URLS.API}tasks/`, data)
     },
 
+    /*---------------------------------------------------------------------
+         Queues
+    ---------------------------------------------------------------------*/
+    get_queues: function (filters) {
+        return CODALAB.api.request('GET', URLS.API + 'queues/', filters)
```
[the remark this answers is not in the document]
Tthomas63: Hadn't heard of template strings before for JS, so that's another tool in the belt. Updated this to use template strings.

src/static/js/ours/client.js
```diff
@@ -237,6 +237,25 @@ CODALAB.api = {
...
+    },
+    get_queue: function (pk) {
+        return CODALAB.api.request('GET', `${URLS.API}queues/${pk}/`)
+    },
+    delete_queue: function (id) {
+        return CODALAB.api.request('DELETE', URLS.API + 'queues/' + id + '/')
```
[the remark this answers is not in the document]
Tthomas63: Updated this to use template strings.

src/static/riot/queues/list.tag
Tthomas63: Talked to you about this, should be addressed

src/static/riot/queues/list.tag
Tthomas63: See above, addressed

src/static/riot/queues/list.tag
Tthomas63: Updated

src/templates/queues/list.html
Tthomas63: Pulled `queues/base_queues.html` and set to extend `base.html`

src/static/riot/queues/list.tag
Tthomas63: Pulled. Previously how this was written I think it was causing the page to refresh was the reason for this.

src/static/riot/queues/list.tag
Tthomas63: I did verify this is working as intended.

src/static/riot/queues/list.tag
Tthomas63: Fixed

src/static/riot/queues/list.tag
Tthomas63: Replaced JQuery references to use `self.refs`

src/static/riot/queues/list.tag
Tthomas63: I did verify pagination and fixed a small issue with it. It should be working as intended.

src/static/riot/queues/list.tag
Tthomas63: Verified this displays correctly without any queues

src/static/riot/competitions/editor/_competition_details.tag
```diff
@@ -25,6 +24,19 @@
             <label>Description</label>
             <textarea class="markdown-editor" ref="comp_description" name="description"></textarea>
         </div>
+        <div class="field">
+            <label>Queue</label>
+            <div class="ui fluid search selection dropdown queue-search">
```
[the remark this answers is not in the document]
Tthomas63: Using `<select></select>` now instead of `<div></div>`

src/apps/api/serializers/queues.py
```diff
@@ -0,0 +1,75 @@
...
+            'vhost',
+            'is_public',
+            'owner',
+            'organizers',
+            'broker_url',
+            'created_when',
+            'is_owner',
+            'id',
+        ]
+        read_only_fields = [
```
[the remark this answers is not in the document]
Tthomas63: Changed

src/apps/api/serializers/queues.py
```diff
@@ -30,28 +32,33 @@ class Meta:
...
+        if request.user.queues.count() == request.user.rabbitmq_queue_limit and not request.user.is_superuser:
+            # TODO: Do we want to raise a validation error (400, bad request) or something that will raise a 403?
+            raise ValidationError("User has reached queue limit!")
+        return super().validate(attrs)
 
     def get_is_owner(self, instance):
-        if instance.owner == self.context.get('owner'):
-            return True
-        return False
+        return instance.owner == self.context.get('owner')
```
jimmykodes: would it be possible for a queue to have no owner, and `context.get('owner')` to also be empty? in which case we would be saying: `return None == None` i.e. `is_owner=True` which we obviously don't wanna do?
Maybe do `return instance.owner and instance.owner == self.context.get('owner')`

src/apps/api/tests/test_queue_api_permissions.py
```diff
@@ -4,15 +4,15 @@
 from django.test import TestCase
 from django.urls import reverse
 
-from factories import UserFactory
+from factories import UserFactory, QueueFactory
 from queues.models import Queue
 
 
 class QueuesAPITestCase(TestCase):
```
jimmykodes: overall, in all these tests, we should probably assert what status codes _are_ rather than what they _are not_. Maybe something changes and something that previously returned a 201 now just returns a 200, and we want it it return a 403. Asserting `response.status_code != 201` would still pass with a `200` response when it should fail.

src/apps/api/tests/test_queue_api_permissions.py
```diff
@@ -104,3 +99,42 @@ def test_organizer_user_cannot_delete_or_edit_queue(self):
...
         assert response.status_code != 204
         assert Queue.objects.filter(pk=self.created_queue.id).exists()
+
+    # Test user queue limits
+    def test_queue_limits_for_regular_users(self):
+        [QueueFactory(owner=self.user) for _ in range(3)]
+
+        self.client.login(username='test', password='test')
+        data = self._get_test_update_data(return_id=False)
+        # default limit is 5
```
jimmykodes: add
`assert self.user.<whatever the path to queue limit is> == 5`
so that this test is more secure.

... or ... you could change the rabbit queue limit to be < 5, then you wouldn't need to do the `QueueFactory() for _ in range()` bit. Either way, we shouldn't build a test around a default value, since maybe that will change, and break our test.

src/apps/api/tests/test_queue_api_permissions.py
```diff
@@ -104,3 +99,42 @@ def test_organizer_user_cannot_delete_or_edit_queue(self):
...
+        # Next one should fail!
+        response = self.client.post(
+            reverse('queues-list'),
+            data=data,
+            content_type='application/json'
+        )
+        assert response.status_code == 400
+
+    # Test user queue limits
+    def test_queue_limits_for_super_users(self):
```
jimmykodes: rename test to something like
`test_super_users_have_no_queue_limit`
since that is what we are actually doing.

src/apps/api/tests/test_queue_api_permissions.py
```diff
@@ -4,15 +4,15 @@
...
 
 
 class QueuesAPITestCase(TestCase):
     def setUp(self):
         self.user = UserFactory(username='test')
         self.another_user = UserFactory(username='another_user')
-        self.organizer_user = UserFactory(username='organizer_user')
+        self.collab_user = UserFactory(username='organizer_user')
         # Prevent queue creation
         with mock.patch('queues.models.rabbit.create_queue') as rabbit_create_queue:
```
jimmykodes: This should get abstracted to a helper function called something like `mock_create_queue` and it should use `QueueFactory()` and then use this function in place of `QueueFactory` so that anywhere below that you are currently using `QueueFactory`, `create_queue` will get mocked correctly every time. (that sentence is a mess... let me know if it doesn't make sense...)

src/apps/api/serializers/queues.py
```diff
@@ -30,28 +32,33 @@ class Meta:
...
             'created_when',
             'is_owner',
-        ]
+        )
+
+    def validate(self, attrs):
+        request = self.context.get('request')
+        if request.user.queues.count() == request.user.rabbitmq_queue_limit and not request.user.is_superuser:
+            # TODO: Do we want to raise a validation error (400, bad request) or something that will raise a 403?
+            raise ValidationError("User has reached queue limit!")
```
jimmykodes: Talked to Eric, we think we should probably raise `PermissionDenied` instead of `ValidationError` here, so the correct error will get sent back to the front end.

src/apps/api/tests/test_queue_api_permissions.py
```diff
@@ -46,12 +56,7 @@ def test_organizer_can_perform_all_operations(self):
         self.client.login(username='test', password='test')
         response = self.client.put(
```
jimmykodes: Do we maybe need to mock things here, too?

src/apps/api/tests/test_queue_api_permissions.py
```diff
@@ -104,3 +99,42 @@ def test_organizer_user_cannot_delete_or_edit_queue(self):
...
+        )
+        assert response.status_code == 400
+
+    # Test user queue limits
+    def test_queue_limits_for_super_users(self):
+        # Make test user a superuser before hand
+        self.user.is_superuser = True
+        self.user.save()
+
+        [QueueFactory(owner=self.user) for _ in range(5)]
```
jimmykodes: given you create a queue in setup already, this should be `range(4)` so that the total is at the limit? or, as I said above, could probably explicitly set the user queue limit lower, so we don't have to run the factory as many times. Will also speed these test up (minimally)

src/static/riot/competitions/editor/_competition_details.tag
```diff
@@ -71,30 +62,23 @@
...
-                    onChange: (value, title) => {
-                        self.refs.queue.value = value
-                        self.form_updated()
-                    }
-                })
-            ;
-
-            self.get_available_queues()
+            $(self.refs.queue).dropdown({
+                // Note: Passing `public=true` so default behavior is users can search for public queues
```
jimmykodes: Does passing `public=True` screw up the `get_queryset` at all? since we list 'public' as one of the filter fields in the `QueueViewSet`, will this filter out all the user's queues that are not set to public? I can't remember the firing order of 'filter_backends' and 'get_queryset'

src/static/riot/competitions/editor/_competition_details.tag
```diff
@@ -71,30 +62,23 @@
...
+                apiSettings: {
+                    url: `${URLS.API}queues/?search={query}&public=true`,
+                },
+                clearable: true,
+                minCharacters: 2,
+                fields: {
+                    remoteValues: 'results',
+                    name: 'name',
```
jimmykodes: if `name` maps to `name`, I don't think we need this line?

src/static/riot/competitions/editor/_competition_details.tag
jimmykodes: Can you make a github issue with information on how to replicate the error you were getting so that we can debug this?

src/static/riot/queues/management.tag
```diff
@@ -80,25 +80,23 @@
...
-                            <label>Is Public?</label>
-                            <input type="checkbox" ref="queue_public">
-                        </div>
+                <div class="required field">
+                    <label>Name</label>
+                    <input name="name" placeholder="Name" ref="queue_name" value="{ _.get(selected_queue, 'name') }">
+                </div>
+                <div class="field">
+                    <div class="ui checkbox">
+                        <label>Is Public?</label>
```
jimmykodes: instead of "Is public?" lets do "Make Public" so that the purpose of the checkbox is more clear.

src/apps/api/tests/test_queue_api_permissions.py
Tthomas63: ✔️

src/apps/api/tests/test_queue_api_permissions.py
Tthomas63: Yes you're right because update will trigger it as well, I created another helper function to make API requests while mocking. Let me know if you think this is gross:
```
    def queue_api_request_with_mock(self, http_method, url_name, url_kwargs, data):
        # So we can make API requests without triggering rabbitmq actual queue creation
        with mock.patch('queues.models.rabbit.create_queue') as rabbit_create_queue:
            rabbit_create_queue.return_value = uuid.uuid4()
            response = getattr(self.client, http_method)(
                reverse(url_name, kwargs=url_kwargs),
                data=data,
                content_type='application/json'
            )
            assert rabbit_create_queue.called
            return response
```
(Having issues with `assert rabbit_create_queue.called` here, still figuring that out)

src/apps/api/tests/test_queue_api_permissions.py
Tthomas63: Changed to just lower user limit.
✔️

src/apps/api/tests/test_queue_api_permissions.py
Tthomas63: ✔️

src/apps/api/tests/test_queue_api_permissions.py
Tthomas63: Set queue limit lower, so this is more explicit, and assert we're already at the limit before making another successful request
✔️

src/static/riot/competitions/editor/_competition_details.tag
Tthomas63: ✔️

src/static/riot/queues/management.tag
Tthomas63: ✔️

src/static/riot/competitions/editor/_competition_details.tag
Tthomas63: This is something I'll have to verify. The behavior when tested previously was working as I wanted.

src/static/riot/competitions/editor/_competition_details.tag
Tthomas63: Can do

src/apps/api/tests/test_queue_api_permissions.py
Tthomas63: Figured that out, and updated everything appropriately. Helper function returns response, and that. In the appropriate tests we assert it was called.

src/static/riot/competitions/editor/_competition_details.tag
Tthomas63: image

Talked to you previously on Slack, I believe this verifies that user's queues that are not public will not be filtered out.

src/static/riot/competitions/editor/_competition_details.tag
Tthomas63: ✔️

src/static/riot/competitions/editor/_competition_details.tag
Tthomas63: Also added a test for this

src/apps/api/tests/test_queues.py
```diff
@@ -0,0 +1,177 @@
...
+            'delete',
+            'queues-detail',
+            {'pk': self.created_queue.id},
+            {}
+        )
+        assert response.status_code == 204
+        assert not Queue.objects.filter(pk=self.created_queue.id).exists()
+
+    def test_other_user_cannot_delete_or_edit_queue(self):
+        self.client.login(username='another_user', password='another_user')
```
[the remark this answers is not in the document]
jimmykodes: image
image

So your tests are passing, but they aren't testing what you think you are testing. looks like you are basically testing permissions of `AnonymousUser`

src/apps/api/tests/test_queues.py
```diff
@@ -0,0 +1,177 @@
...
+            {}
+        )
+
+        assert response.status_code == 403
+        assert Queue.objects.filter(pk=self.created_queue.id).exists()
+
+    # --------- Queue Limit Tests ---------
+
+    # Test user queue limits
+    def test_queue_limits_for_regular_users(self):
```
jimmykodes: You didn't log in here?

src/apps/api/tests/test_queues.py
```diff
@@ -0,0 +1,177 @@
...
+            self._get_test_update_data(return_id=False)
+        )
+        assert response.status_code == 403
+        assert self.user.rabbitmq_queue_limit == self.user.queues.count()
+
+    # Test super user queue limits
+    def test_super_users_has_no_queue_limit(self):
+        # Make test user a superuser before hand
+        self.user.is_superuser = True
+        self.user.rabbitmq_queue_limit = 2
```
jimmykodes: you can set these values in the UserFactory, so you don't need to do it in the tests

src/apps/api/tests/test_queues.py
```diff
@@ -0,0 +1,177 @@
...
+        self.mock_create_queue(owner=self.another_user)
+
+        # Create yet another queue that IS public, that we want filtered out by search
+        self.mock_create_queue(owner=self.another_user, is_public=True)
+
+        self.client.login(username='test', password='test')
+
+        response = self.client.get(f"{reverse('queues-list')}?search=test&public=True")
+        assert response.status_code == 200
+        assert len(response.json()['results']) == 2
```
jimmykodes: response.data not response.json

src/apps/competitions/tasks.py
```diff
@@ -338,6 +361,17 @@ def unpack_competition(competition_dataset_pk):
                 "solutions": {},
             }
 
+            # Get Queue by vhost/uuid. If instance not returned, or we don't have access don't set it!
+            if competition_yaml.get('queue'):
+                try:
+                    queue = Queue.objects.get(vhost=competition_yaml.get('queue'))
+                    if not queue.is_public:
+                        if queue.owner != creator or creator.username not in queue.organizers.all().values_list('username'):
```
jimmykodes: `values_list('username', flat=True)` is probably more what ya want, otherwise:
image

src/static/riot/queues/management.tag
```diff
@@ -0,0 +1,271 @@
...
+            navigator.clipboard.writeText(queue.broker_url).then(function () {
+                // clipboard successfully set
+                toastr.success("Successfully copied broker url to clipboard!")
+            }, function () {
+                // clipboard write failed
+                toastr.error("Failed to copy broker url to clipboard!")
+            });
+        }
+    </script>
+    <style>
```
jimmykodes: use stylus.

jimmykodes:
image
It looks like the copy button in the queue management section doesn't work, which means I don't have any way to grab the broker URL. Since the eyeball button is a tooltip, I can see the url, but I cannot hover over it to copy it. We should also/instead display the URL some other way, just in case this copy button doesn't work in other browsers or something perhaps?

jimmykodes:
Looks like sending submissions to the new broker URL is also working, so fix those small UI bugs, and I think this is good to go.

PR #249, opened 2019-11-06
Front page changes

gibsonbailey:


src/apps/api/views/competitions.py
```diff
@@ -21,6 +21,7 @@
 
 class CompetitionViewSet(ModelViewSet):
     queryset = Competition.objects.all()
+    permission_classes = (AllowAny,)
```
ckcollab: does this let anyone delete a competition, as well?

src/factories.py
```diff
@@ -12,6 +14,11 @@
 from tasks.models import Task
 
 
+def _get_random_date():
```
ckcollab: nice

src/static/riot/display_statistics/show_stats.tag
```diff
@@ -1,26 +1,13 @@
...
-                        <div class="statistic">
-                            <div class="value">
-                                { stat.count }
-                            </div>
-                            <div class="label">
-                                { stat.label }
-                            </div>
-                        </div>
+    <div class="ui six column grid">
+        <div class="column" each="{ stat in general_stats }" no-reorder>
```
ckcollab: thank you!

ckcollab:
Merged in develop, hopefully tests pass

src/factories.py
gibsonbailey: This is the cause for the failing test. I will have to look into it.

src/static/riot/competitions/detail/_tabs.tag
```diff
@@ -314,7 +314,9 @@
                     })
                 })
             })
-            self.selected_leaderboard_index = self.competition.leaderboards[0].id
+            if (self.competition.leaderboards.length > 0) {
+                self.selected_leaderboard_index = self.competition.leaderboards[0].id
```
jimmykodes: could do this a couple different ways:
```javascript
self.selected_leaderboard_index = _.get(_.head(self.competition.leaderboards), 'id')
```
or
```javascript
if (!_.isEmpty(self.competition.leaderboards)) {
    self.selected_leaderboard_index = self.competition.leaderboards[0].id
}
```

src/static/riot/display_statistics/show_stats.tag
jimmykodes: should this be: 
```html
<div class="ui six small statistics">
    <virtual each="{stat in general_stats}">
        <div class="statistic">
            <div class="value">
                { stat.count }
            </div>
            <div class="label">
                { stat.label }
            </div>
        </div>
    </virtual>
</div>
```
instead of putting each stat in its own column, let semantic handle the spacing?

src/templates/base.html
```diff
@@ -223,6 +242,22 @@ <h4 class="ui inverted header">Footer Header</h4>
         }
         // Django returns <input type="hidden" value="csrf-token-secret"> so we just grab the `value` attribute from this
         CSRF_TOKEN = $('{% csrf_token %}').attr('value')
+
+        $(document).ready( function () {
+            $('.ui.search').search({
```
jimmykodes: make this jquery selector more specific (use an `#id` probably?) so we don't conflict with anything in `.tag` fils.

src/templates/base.html
```diff
@@ -223,6 +242,22 @@ <h4 class="ui inverted header">Footer Header</h4>
...
+                apiSettings: {
+                    onResponse: function(codalabResponse) {
+                        _.forEach(codalabResponse, (response) => {
+                            response.url = '/competitions/' + response.id
```
jimmykodes: Should be
`response.url = URLS.COMPETITION_DETAIL(response.id)` so that this isn't hardcoded.

src/templates/base.html
```diff
@@ -223,6 +242,22 @@ <h4 class="ui inverted header">Footer Header</h4>
...
+                        })
+                        return {
+                            results: codalabResponse,
+                        }
+                    },
+                    url: URLS.API + 'competitions/?search={query}'
```
jimmykodes:
```javascript
url: `${URLS.API}competitions/?search={query}`
```

src/templates/base.html
```diff
@@ -240,4 +275,14 @@ <h4 class="ui inverted header">Footer Header</h4>
     {% endblock %}
 
 </body>
+<style>
```
jimmykodes: can we do this in a stylus file somewhere? so we don't have to have this css in base.html?

src/factories.py
jimmykodes: This is built in to factoryboy via faker, so you don't need this helper function.
example:
https://github.com/codalab/chahub/blob/e66203d36ae43802b2f304a0564c941bc9b1e174/src/factories.py#L148

jimmykodes: (to fix tests, you'll probably just need to add `created_when=timezone.now()` to the test (maybe a helper function) that is creating the submissions)

src/static/riot/display_statistics/show_stats.tag
ckcollab: I misread this as `{{ stat in general_stats }}`, ughhhh, this should be reading from Django Context Processor not from an API endpoint

src/factories.py
gibsonbailey: Great call, Jim. Much cleaner! ✅

src/static/riot/display_statistics/show_stats.tag
gibsonbailey: Semantic is deciding to throw the last stat on a new line if I don't mandate that they're columns. 

This has been moved to home.html and the data is provided by a context processor.

src/static/riot/competitions/detail/_tabs.tag
ckcollab: ✔️

src/templates/base.html
ckcollab: ✔️

src/apps/api/views/competitions.py
gibsonbailey: ✅

PR #272, opened 2019-11-26
Task solution validation

ckcollab:


docker/compute_worker/compute_worker.py
```diff
@@ -86,6 +87,15 @@ def replace_legacy_metadata_command(command, kind, is_scoring, ingestion_only_du
     return command
 
 
+def md5(filename):
```
ckcollab: I duplicate this below.. :\

ckcollab:
TODO: Make frontend have a `public` button of some kind..

src/tests/functional/test_submissions.py
```diff
@@ -48,19 +52,37 @@ def _run_submission(self, competition_zip_path, submission_zip_path, expected_su
         # The submission table lists our submission!
         assert self.find('submission-manager table tbody tr:nth-of-type(1) td:nth-of-type(2)').text == submission_zip_path
 
+        # Check that md5 information was stored correctly
+        # TODO: Why in @$#! does this not work:
+        #   os.path.join("./src/tests/functional", submission_full_path)
```
jimmykodes: this is because `submission_full_path` is an absolute path, (it starts with a `/`) so `os.path.join()` can't join an absolute path on to the end of a different one. sooo, you could strip the leading `/` off the `submission_full_path` and it works:
```python
    def test_md5(self):
        submission_full_path = os.path.join(self.test_files_dir, 'submission_15.zip')
        f_string = md5(f"./src/tests/functional{submission_full_path}")
        join_path = md5(os.path.join('./src/tests/functional', submission_full_path[1:]))
        assert f_string == join_path
```
Not sure that is a _great_ solution, but at least we know why it doesn't work now...

src/apps/api/urls.py
```diff
@@ -19,7 +19,8 @@
 router.register('leaderboards', leaderboards.LeaderboardViewSet)
 router.register('submission_scores', leaderboards.SubmissionScoreViewSet, 'submission_scores')
 router.register('tasks', tasks.TaskViewSet)
-router.register('tasksearch', tasks.TaskViewSetSimple)
+# TODO: We don't need a whole separate viewset to filter tasks, also messes up reverse searching 'task-list'
+# router.register('tasksearch', tasks.TaskViewSetSimple)
```
jimmykodes: This is used for searching tasks on comp editor phase modal. Removing this viewset will break that, we could maybe implement this behavior on an action? but the easier fix (for now) is to add a basename to the router.register:
```python
router.register('tasksearch', tasks.TaskViewSetSimple, base_name='tasksearch')
```
and then, at the very least, you won't get `reverse` issues, and task searching will still work.

src/factories.py
```diff
@@ -84,6 +84,22 @@ class Meta:
     name = factory.Sequence(lambda n: f'Task {n}')
     created_by = factory.SubFactory(UserFactory)
 
+    @factory.post_generation
+    def solutions(self, create, extracted, **kwargs):
+        if not create:
+            return
+
+        if extracted:
```
jimmykodes: slight improvement to this factory could be like we do on `SubissionScoreFactory` and check to see if `isinstance(extracted, Solution)` and then `self.solutions.add(extracted)` else do the `for solution in extracted` so that we can pass either a single solution object, _or_ a list of them to this factory.

src/tests/functional/test_submissions.py
```diff
@@ -48,19 +52,37 @@ def _run_submission(self, competition_zip_path, submission_zip_path, expected_su
...
+        #   it returns..
+        #   submission_full_path, every time...
+        submission_md5 = md5(f"./src/tests/functional{submission_full_path}")
+        assert Submission.objects.get(md5=submission_md5)
+        for solution in Solution.objects.all():
+            print(f"md5: {solution.md5}, path {solution.data.data_file.name}")
+        assert Solution.objects.get(md5=submission_md5)
```
jimmykodes: same here, `assert exists()`

docker/compute_worker/compute_worker.py
```diff
@@ -421,11 +444,20 @@ def prepare(self):
 
         if self.is_scoring:
             # Send along submission result so scoring_program can get access
-            bundles += [(self.prediction_result, os.path.join('input', 'res'))]
+            bundles += [(self.prediction_result, 'input/res')]
```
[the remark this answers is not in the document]
ckcollab: when I was searching for `input/res` I could not find this, and it matches the same way I typed `input/ref` above :\ maybe not the best reasoning, but the consistency may help someone 1 year from now grok this

src/apps/api/urls.py
ckcollab: I feel like a normal `GET` to `/tasks` should be the "search?" not necessary to have an action or something?

src/factories.py
ckcollab: I think we should have a convention where we just always pass an array to this? hmm... 

that way we don't have to modify _every_ factory, we just know for m2m/whatever we pass `[single_item]`

src/tests/functional/test_submissions.py
ckcollab: ughhhhhhhhhh, how have I not ran into this after so many years? Annoying. ty!

Pulling this comment and leaving in this:

```md5(f"./src/tests/functional{submission_full_path}")```

File: src/tests/functional/test_submissions.py [hunk 7]
```diff
@@ -48,19 +52,37 @@ def _run_submission(self, competition_zip_path, submission_zip_path, expected_su
         # The submission table lists our submission!
         assert self.find('submission-manager table tbody tr:nth-of-type(1) td:nth-of-type(2)').text == submission_zip_path
 
+        # Check that md5 information was stored correctly
+        # TODO: Why in @$#! does this not work:
+        #   os.path.join("./src/tests/functional", submission_full_path)
+        #   it returns..
+        #   submission_full_path, every time...
+        submission_md5 = md5(f"./src/tests/functional{submission_full_path}")
+        assert Submission.objects.get(md5=submission_md5)
```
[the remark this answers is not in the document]
ckcollab: tyty

src/tests/functional/test_submissions.py
ckcollab: ty

src/apps/api/urls.py
jimmykodes: they require different serializers, how are we going to pass a different serializer for the different get requests?

ckcollab: I believe we could make a "simple list serializer" that works for both a simple "get" and a "get with search params?" It may return a bit more data than you need in some cases, but that should be fine?

src/factories.py
jimmykodes: 🤷‍♂ yea, phases is that way, always have to pass a list of tasks. Does seem weird to always have to remember to wrap it in a list, since most of the time w/ a factory, we are probably only going to be passing one element anyway...

ckcollab: Yeah it is kind of weird, it's the same way with sending emails `send_mail(from=one_email, to=[list_of_emails])` even though I'm usually just sending an email to one person, that's how the interface works =\

I'll think about this a bit more.. hmm...

src/apps/api/urls.py
jimmykodes: I guess, though we are changing the names of one of the fields in the simple serializer `key -> value` so that it will work w/ the multiselect. I guess we could just return the key twice? once as key, once as value?

src/factories.py
ckcollab: Also, in this case we're passing to a "plural variable:"

```
task = TaskFactory(created_by=user, solutions=[solution])

...

phase = PhaseFactory(competition=competition, tasks=[task])
```

I _think_ this pattern makes sense..

PR #274, opened 2019-12-03
better unpacking exception handling

### LINKED ISSUE #273: Bundle uploading/unpacking taking forever
I'm trying to upload the following bundle to Codalab V2 (prod LRI instance):

[Mortality challenge bundle.](https://drive.google.com/open?id=1fPiNkU-tn-2awkPGWsADY4yuV63QZLSU)

The uploading seems to work but then the **unpacking** step never finishes.

ckcollab:
fixes #273 

# Example

render of the following error dict:

```
{
    "pages": {
        "content": ["This field may not be blank."]
    }
}
```

image

# known issues

may not make 100% sense to a v1.5 person as the errors correspond more directly to V2 YAML =\

### DIFF
--- a/src/apps/competitions/tasks.py
+++ b/src/apps/competitions/tasks.py
@@ -305,12 +305,21 @@ def unpack_competition(competition_dataset_pk):
             try:
                 competition = unpacker.save()
             except ValidationError as e:
-                error_str = ""
-                for key, errors in e.detail.items():
-                    error_str += f'{key}: {"; ".join(errors)}\n'
-                raise CompetitionUnpackingException(
-                    f"Bundle configuration errors. {error_str}"
-                )
+                def _get_error_string(error_dict):
+                    """Helps us nicely print out a ValidationError"""
+                    for key, errors in error_dict.items():
+                        try:
+                            return f'{key}: {"; ".join(errors)}\n'
+                        except TypeError:
+                            # We ran into a list of nested dictionaries, start recursing!
+                            nested_errors = []
+                            for e in errors:
+                                error_text = _get_error_string(e)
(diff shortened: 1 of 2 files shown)


src/apps/competitions/tasks.py
```diff
@@ -299,20 +301,47 @@ def unpack_competition(competition_dataset_pk):
...
         status.details = str(e)
         status.status = CompetitionCreationTaskStatus.FAILED
         status.save()
         raise e
 
+    except Exception as e:
+        # These are critical uncaught exceptions, make sure the end user is at least informed
+        # that unpacking has failed -- do not share unhandled exception details
+        logger.error(traceback.format_exc())
+        status.details = "Contact an administrator, competition failed to unpack in a critical way."
```
jimmykodes: Nice.

src/apps/competitions/unpackers/v1.py
```diff
@@ -7,11 +7,17 @@
 class V15Unpacker(BaseUnpacker):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+
+        # Just in case docker image is blank (""), replace with default value
+        docker_image = self.competition_yaml.get('competition_docker_image', 'codalab/codalab-legacy:py3')
```
jimmykodes: maybe do:
```python
docker_image = self.competition_yaml.get('competition_docker_image')
if not docker_image:
    docker_image = "codalab/codalab-legacy:py3"
```
so that we replace _every_ falsy value?

ckcollab: I'll switch to this style

PR #298, opened 2020-01-28
Misc v2 fixes (duplicate participant, non-base64 logos, "none" user lookup)

ckcollab:
# @ mention of reviewers
@jimmykodes 

Make CompetitionParticipant unique on user and competition

- [x] can participate in competition
- [x] can't participate multiple times
- [x] can add a collaborator after they've requested participation and things still work
- [x] also fixes bug with newly created user having a "None" name (should also work on Queue's page)
- [x] Changing logos in a competition should still work

# Extra credit

- [x] Add a competition test confirming a participant gets "accepted" if added as organizer and they are pending participation acceptance _(thinking it is OK to skip acceptance email?)_

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge


src/apps/api/tests/test_competition_participants.py
```diff
@@ -46,3 +48,13 @@ def test_normal_user_cannot_see_participants(self):
         resp = self.get_participants()
         assert CompetitionParticipant.objects.filter(competition=self.comp).count() > 0
         assert len(resp.json()) == 0
+
+    def test_cant_participate_multiple_times(self):
+        self.client.login(username='norm', password='norm')
+        resp = self.client.post(reverse('competition-register', kwargs={"pk": self.comp.pk}))
+        assert resp.status_code == 201
+        CompetitionParticipant.objects.get(user=self.norm)
```
jimmykodes: I assume this is verifying that the CompetitionParticipant was created? probably better to do something like 
```suggestion
        assert CompetitionParticipant.objects.filter(user=self.norm, competition=self.comp).exists()
```
so we don't get a `CompetitionParticipant.DoesNotExist` error in tests

src/apps/api/tests/test_competition_participants.py
```diff
@@ -46,3 +48,13 @@ def test_normal_user_cannot_see_participants(self):
...
+        resp = self.client.post(reverse('competition-register', kwargs={"pk": self.comp.pk}))
+        assert resp.status_code == 400
+        assert "You already applied for participation in this competition!" in resp.data[0]
+        assert len(CompetitionParticipant.objects.filter(user=self.norm)) == 1
```
jimmykodes: for better optimization and accuracy
```suggestion
        assert CompetitionParticipant.objects.filter(user=self.norm, competition=self.comp).count() == 1
```

src/apps/api/tests/test_competitions.py
```diff
@@ -0,0 +1,52 @@
...
+
+        # Just get the key from the task and pass that instead of the object
+        data["phases"][0]["tasks"] = [data["phases"][0]["tasks"][0]["key"]]
+        return data
+
+    # TODO: Do we have competition permissions tests?
+    # def test_cant_edit_someone_elses_competition?
+
+    def test_adding_organizer_accepts_them_as_participant(self):
+        CompetitionParticipantFactory(
```
jimmykodes: the user shouldn't need to already have a "pending" status for them to be added as a participant. should maybe test both use cases? that they get added if they haven't already applied and then accepted if they have.

src/apps/api/tests/test_competitions.py
```diff
@@ -0,0 +1,52 @@
...
+        self.client.login(username='creator', password='creator')
+        url = reverse('competition-detail', kwargs={"pk": self.comp.pk})
+
+        # Get comp data to work with
+        data = self._prepare_competition_data(url)
+
+        data["collaborators"] = [self.other_user.pk]
+        resp = self.client.put(url, data=json.dumps(data), content_type="application/json")
+        assert resp.status_code == 200
+        assert CompetitionParticipant.objects.get(user=self.other_user, competition=self.comp, status=CompetitionParticipant.APPROVED)
```
jimmykodes: I always try to avoid using `get` in tests and assert `exists` or `count` instead, because if we break this down the road, we get a nice pytest failure message, instead of a django `ObjectDoesNotExist` error.

src/apps/api/views/competitions.py
```diff
@@ -169,6 +174,19 @@ def email_all_participants(self, request, pk):
         batch_send_email.apply_async((comp.pk, content))
         return Response({}, status=status.HTTP_200_OK)
 
+    def _insure_organizer_participants_accepted(self, instance):
```
jimmykodes: should be `ensure`

src/static/js/ours/utils.js
```diff
@@ -117,14 +114,36 @@ function set_form_data(data, base_element) {
...
-            //console.log("@@@@@")
-            //console.log(field)
-            console.log(field.name + " -> " + data[field.name])
             $(field).val(data[field.name])
         }
     })
 }
 
+function objectToFormData(obj, form, namespace) {
+    var fd = form || new FormData()
```
jimmykodes: use `let` or `const`, not `var`

src/static/js/ours/utils.js
```diff
@@ -117,14 +114,36 @@ function set_form_data(data, base_element) {
...
+    var formKey
+
+    for (var property in obj) {
```
jimmykodes: having trouble wrapping my head around this entirely (it has been a long day...) but could we be doing this with something like `_.reduce`? this kinda seems like a reduce usecase? or am I confused? would be very on board with some tag-team code golfing this. (or maybe just an explanation of what I'm looking at....)

src/static/riot/queues/management.tag
```diff
@@ -186,10 +184,12 @@
         Modal Methods
         ----------------------------------------------------------------------*/
         self.show_modal = (queue) => {
+            // Show modal THEN update!
+            $(self.refs.modal).modal('show')
```
jimmykodes: Why does this matter? what changes between doing this before vs after?

src/tests/functional/test_competitions.py
```diff
@@ -27,14 +27,16 @@ def _upload_competition(self, competition_zip_path):
...
-        task = comp.phases.first().tasks.first()
         created_items = [
             comp.bundle_dataset.data_file.name,
             comp.logo.name,
-            task.scoring_program.data_file.name,
-            task.reference_data.data_file.name,
         ]
+        for phase in comp.phases.all():
+            for task in phase.tasks.all():
+                created_items += [
```
jimmykodes: neat! I always forget you can do this

src/apps/api/tests/test_competitions.py
```diff
@@ -0,0 +1,52 @@
...
+    def setUp(self):
+        self.creator = UserFactory(username='creator', password='creator')
+        self.other_user = UserFactory(username='other_user', password='other')
+        self.comp = CompetitionFactory(created_by=self.creator)
+        PhaseFactory(competition=self.comp)
+        self.leaderboard = LeaderboardFactory(competition=self.comp)
+        ColumnFactory(leaderboard=self.leaderboard)
+
+    def _prepare_competition_data(self, url):
+        # data = CompetitionSerializer(comp).data
```
jimmykodes: pull

src/apps/api/tests/test_competitions.py
```diff
@@ -0,0 +1,52 @@
...
+        data = resp.data
+        data.pop('id')
+        data.pop('logo')
+
+        # Just get the key from the task and pass that instead of the object
+        data["phases"][0]["tasks"] = [data["phases"][0]["tasks"][0]["key"]]
+        return data
+
+    # TODO: Do we have competition permissions tests?
+    # def test_cant_edit_someone_elses_competition?
```
jimmykodes: to my knowledge, we don't have any tests that explicitly verify this code is working
https://github.com/codalab/competitions-v2/blob/935a4df38a94bee22b65214bf3a935eae05229c1/src/apps/api/views/competitions.py#L80

src/apps/api/tests/test_competition_participants.py
ckcollab: ✔️

src/apps/api/tests/test_competition_participants.py
ckcollab: ✔️

src/apps/api/tests/test_competitions.py
ckcollab: ✔️ (note, left this in so we could discuss this way vs doing a `GET` to get competition details)

src/apps/api/tests/test_competitions.py
ckcollab: ✔️ made #299 for this

src/apps/api/tests/test_competitions.py
ckcollab: ✔️

src/apps/api/views/competitions.py
ckcollab: Oof, I was right the first time I wrote it then, ty

✔️

src/static/js/ours/utils.js
ckcollab: ✔️

src/static/js/ours/utils.js
ckcollab: I don't know about a `_.reduce` but to be clear: this is from a gist for turning a nested object into a `FormData` object

src/static/riot/queues/management.tag
ckcollab: I was trying to get this dropdown to work and part of that was thinking this modal needed to be shown first before the dropdown would refresh properly with the selected organizers.

image


good catch, I tested and this was not part of what made it start working, ty!

✔️

src/tests/functional/test_competitions.py
ckcollab: so handy :)

src/apps/api/tests/test_competitions.py
ckcollab: added a test for this other case as well ✔️

ckcollab:
Hand tested, ready for further review! (apologies for the extra iterations on this guy)

src/apps/api/tests/test_competitions.py
```diff
@@ -0,0 +1,63 @@
...
+        data["logo"] = None
+
+        # Just get the key from the task and pass that instead of the object
+        data["phases"][0]["tasks"] = [data["phases"][0]["tasks"][0]["key"]]
+        return data
+
+    # TODO: Do we have competition permissions tests?
+    # def test_cant_edit_someone_elses_competition?
+
+    def test_adding_organizer_accepts_creates_accepted_participant(self):
```
jimmykodes:
```suggestion
    def test_adding_organizer_creates_accepted_participant(self):
```

src/apps/api/tests/test_competitions.py
```diff
@@ -0,0 +1,63 @@
...
+        self.client.login(username='creator', password='creator')
+        url = reverse('competition-detail', kwargs={"pk": self.comp.pk})
+
+        # Get comp data to work with
+        data = self._prepare_competition_data(url)
+
+        data["collaborators"] = [self.other_user.pk]
+        resp = self.client.put(url, data=json.dumps(data), content_type="application/json")
+        assert resp.status_code == 200
+        assert CompetitionParticipant.objects.filter(user=self.other_user, competition=self.comp, status=CompetitionParticipant.APPROVED).count() == 1
```
jimmykodes: Break this line up maybe?
```suggestion
        assert CompetitionParticipant.objects.filter(
            user=self.other_user, 
            competition=self.comp,
            status=CompetitionParticipant.APPROVED
        ).count() == 1
```

src/apps/api/tests/test_competitions.py
```diff
@@ -0,0 +1,63 @@
...
```
jimmykodes: same here? maybe break up the long line?

src/apps/api/views/competitions.py
```diff
@@ -103,6 +104,36 @@ def get_serializer_context(self):
...
+        self.perform_create(serializer)
+        headers = self.get_success_headers(serializer.data)
+
+        # Re-do serializer in detail version (i.e. for Collaborator data)
+        serializer = CompetitionDetailSerializer(serializer.instance)
+        return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
+
+    def update(self, request, *args, **kwargs):
+        """Mostly a copy of the underlying base update, however we return some additional data
+                in the response to remove a GET from the frontend"""
```
jimmykodes: extra whitespace

src/apps/api/tests/test_competitions.py
ckcollab: ✔️

src/apps/api/tests/test_competitions.py
ckcollab: ✔️

src/apps/api/tests/test_competitions.py
ckcollab: ✔️

src/apps/api/views/competitions.py
ckcollab: ✔️

PR #320, opened 2020-02-10
Fix Heroku deploys

ckcollab:
# @ mention of reviewers
@jimmykodes 

Fixes Heroku (only used for review apps atm)

- [x] The review app for this PR works (https://competitions-v2-staging-pr-320.herokuapp.com/)
- [x] `generate_data` was called on review app creation
- [x] a competition in general works on the 

# Known issues
Generate_data'd competitions won't work? they aren't "real competitions?"

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/src/apps/commands/management/commands/generate_data.py
+++ b/src/apps/commands/management/commands/generate_data.py
@@ -1,10 +1,14 @@
 import random
 
+from django.contrib.auth import get_user_model
 from django.core.management.base import BaseCommand
 from django.db import IntegrityError
 
 from factories import UserFactory, CompetitionFactory, PhaseFactory, SubmissionFactory, CompetitionParticipantFactory, \
     TaskFactory
+
+
+User = get_user_model()
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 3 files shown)


src/apps/commands/management/commands/generate_data.py
```diff
@@ -28,7 +32,20 @@ def handle(self, *args, **kwargs):
         size = kwargs.get('size') or 3
         no_admin = kwargs.get('no_admin')
         print(f'Creating data of size {size} {"without an admin account." if no_admin else "with an admin account." }')
-        users = [UserFactory(username='admin', super_user=True) if i == 0 and not no_admin else UserFactory() for i in range(size)]
+        users = []
+        for i in range(size):
+            if i == 0 and not no_admin:
+                try:
+                    user = UserFactory(username='admin', super_user=True)
+                    user.set_password('admin')
```
jimmykodes: you can pass password to the user factory so you don't need to call set password. and the standard password is test, do we need to change that? if so, could just do it on the factory's post generation hook

ckcollab: ✔️

ckcollab: _(keeping password as admin//admin -- can do admin//test if we all prefer that?)_

PR #325, opened 2020-02-14
Detailed results

jimmykodes:
# @ mention of reviewers
@ckcollab 

changes the detailed results output to be 1.8 style, using html pages, rather than chart js.
Other various editor and comp unpacker fixes

Heroku Test server not currently working... not sure why
- [x] should be able to upload attached comp and make the attached submission against it.
- [x] submission should run correctly and the matplotlib graphs from the backend should update in the submission output accordion
- [x] if you refresh the page, the graphs should correctly return

Competition:
[old_autodl.zip](https://github.com/codalab/competitions-v2/files/4216003/old_autodl.zip)

Submission:
[autodl_sub.zip](https://github.com/codalab/competitions-v2/files/4215754/autodl_sub.zip)

# Checklist
- [x] Code review by me 
- [x] Hand tested by me (on local machine, not heroku)
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge


docker/compute_worker/compute_worker.py
```diff
@@ -371,6 +374,21 @@ def _get_bundle(self, url, destination):
 
             # Input from submission (or submission + ingestion combo)
             docker_cmd += ['-v', f'{self.input_dir}:/app/input']
+            if self.detailed_results_url and self.is_scoring and kind == 'program':
+                # we have a detailed results url, we are in the scoring step, and we aren't the ingestion program
+                detail_path = os.path.join(self.output_dir, "detailed_results.html")
+
+                # Is there a better way to create the entire path without having to mkdir each incremental step?
```
ckcollab: If I'm understanding, this may help?

https://docs.python.org/3/library/os.html#os.makedirs

> Recursive directory creation function. Like mkdir(), but makes all intermediate-level directories needed to contain the leaf directory.

...neat!

docker/compute_worker/compute_worker.py
```diff
@@ -371,6 +374,21 @@ def _get_bundle(self, url, destination):
...
+                if not os.path.exists(detail_path):
+                    if not os.path.exists(self.output_dir):
+                        os.mkdir(self.output_dir)
+                    open(detail_path, 'a').close()
```
ckcollab: this makes the file so you can chmod it, I assume? could maybe make this more clear like:

```python
# Create detailed file and set perms so we can read the file to send to the sas URL
open(detail_path, 'a').close()
os.chmod(detail_path, 0o777)
```

docker/compute_worker/compute_worker.py
```diff
@@ -371,6 +374,21 @@ def _get_bundle(self, url, destination):
...
+
+                    # Change perms so we can read the file to send to the sas URL
+                    os.chmod(detail_path, 0o777)
+
+                _command = ['./watch.sh', self.detailed_results_url, detail_path, self.websocket_url, self.output_dir]
+                self.watcher = subprocess.Popen(_command, stderr=subprocess.STDOUT)
```
ckcollab: does `stderr` actually work here?

docker/compute_worker/put_detailed_results.py
```diff
@@ -0,0 +1,45 @@
+import argparse
+import asyncio
+import json
+
+import requests
+import websockets
+
+parser = argparse.ArgumentParser()
```
ckcollab: nice

docker/compute_worker/put_detailed_results.py
```diff
@@ -0,0 +1,45 @@
...
+    help='URL for websocket to communicate with the frontend about updates to the detailed results'
+)
+
+args = parser.parse_args()
+
+
+async def update_frontend():
+    async with websockets.connect(args.websocket_url) as websocket:
+        await websocket.send(json.dumps({
+            "kind": 'detailed_result',
```
ckcollab: This could maybe be named `detailed_result_update_sent` or something?

docker/compute_worker/watch.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+watchmedo shell-command --patterns="*.html" \
```
ckcollab: How often will this update? Doing `watchmedo shell-command --help` I was able to see some "hidden flags" (not on their docs anywhere that I could see..):

```
❯ watchmedo shell-command --help
usage: watchmedo shell-command [-h] [-c COMMAND] [-p PATTERNS]
                               [-i IGNORE_PATTERNS] [-D] [-R]
                               [--interval TIMEOUT] [-w] [-W]
                               [--debug-force-polling]
                               [directories [directories ...]]

    Subcommand to execute shell commands in response to file system events.

    :param args:
        Command line argument options.
    

positional arguments:
  directories           directories to watch (default: '.')

optional arguments:
  -h, --help            show this help message and exit
  -c COMMAND, --command COMMAND
                        shell command executed in response to matching events.
                        These interpolation variables are available to your
                        command string:: ${watch_src_path} - event source
                        path; ${watch_dest_path} - event destination path (for
                        moved events); ${watch_event_type} - event type;
                        ${watch_object} - ``file`` or ``directory`` Note::
                        Please ensure you do not use double quotes (") to
                        quote your command string. That will force your shell
                        to interpolate before the command is processed by this
                        subcommand. Example option usage:: --command='echo
                        "${watch_src_path}"' (default: -)
  -p PATTERNS, --pattern PATTERNS, --patterns PATTERNS
                        matches event paths with these patterns (separated by
                        ;). (default: '*')
  -i IGNORE_PATTERNS, --ignore-pattern IGNORE_PATTERNS, --ignore-patterns IGNORE_PATTERNS
                        ignores event paths with these patterns (separated by
                        ;). (default: '')
  -D, --ignore-directories
                        ignores events for directories (default: False)
  -R, --recursive       monitors the directories recursively (default: False)
  --interval TIMEOUT, --timeout TIMEOUT
                        use this as the polling interval/blocking timeout
                        (default: 1.0)
  -w, --wait            wait for process to finish to avoid multiple
                        simultaneous instances (default: False)
  -W, --drop            Ignore events that occur while command is still being
                        executed to avoid multiple simultaneous instances
                        (default: False)
  --debug-force-polling
                        [debug] forces polling (default: False)
```

I think we want _at least_ `--wait` here so we don't do this multiple times at once, potentially having an old state override a new state. `--drop` would be kind of lame to miss a state change..

ckcollab: There's an issue here for adding a `--debounce` .. maybe something we could do at some point if we're feeling crazy :) or if @gibsonbailey's brothers want an open source project

https://github.com/gorakhargosh/watchdog/issues/315

docker/compute_worker/compute_worker.py
jimmykodes: not quite what I need. since I just need to make a file one folder in. I would still need to `mkdirs()` and `open().close()`

docker/compute_worker/compute_worker.py
```diff
@@ -151,7 +152,40 @@ def __init__(self, run_args):
...
+        while self.watch:
+            if os.path.exists(file_path):
+                new_time = os.path.getmtime(file_path)
+                if new_time != last_modified_time:
+                    last_modified_time = new_time
+                    await self.send_file(file_path)
+                await asyncio.sleep(5)
+            else:
+                await asyncio.sleep(5)
+        else:
```
ckcollab: cool use of while .. else :)

docker/compute_worker/compute_worker.py
```diff
@@ -151,7 +152,40 @@ def __init__(self, run_args):
...
+        last_modified_time = None
+        while self.watch:
+            if os.path.exists(file_path):
+                new_time = os.path.getmtime(file_path)
+                if new_time != last_modified_time:
+                    last_modified_time = new_time
+                    await self.send_file(file_path)
+                await asyncio.sleep(5)
+            else:
+                await asyncio.sleep(5)
```
ckcollab: could do...

```python

            if os.path.exists(file_path):
                new_time = os.path.getmtime(file_path)
                if new_time != last_modified_time:
                    last_modified_time = new_time
                    await self.send_file(file_path)

            await asyncio.sleep(5)
```

(since you sleep in both sides of the conditional)

docker/compute_worker/compute_worker.py
```diff
@@ -151,7 +152,40 @@ def __init__(self, run_args):
...
+        else:
+            # make sure we always send the final version of the file
+            await self.send_file(file_path)
+
+    async def send_file(self, file_path):
```
ckcollab: since this is specific to detailed results, a better name may be `send_detailed_results`? 

and, I believe this could use `_put_file`?

docker/compute_worker/compute_worker.py
```diff
@@ -290,6 +323,9 @@ def _get_bundle(self, url, destination):
                     self._put_file(value["location"], raw_data=value["data"])
 
             logger.info("Program finished")
+            if self.detailed_results_url and kind == 'program':
+                # Only the scoring program should be able to tell the watcher we are done.
```
ckcollab: ty for comment

PR #354, opened 2020-02-25
Adds fabric worker mgmt

### LINKED ISSUE #340: Make fabric/some scripts to update all workers at once
Could use Ansible instead of Fabric, maybe?

ckcollab:
# @ mention of reviewers
@jimmykodes 

Resolves #340 adding a script to maintain compute workers

- [x] prereq: `pip install fab-classic` outside of docker, uses your SSH key on host machine
- [x] create `server_config.yaml` and add the gpus to it (I'll just send you this), run `fab -R autodl-gpu status`
- [ ] update some workers with `fab -R autodl-gpu update` _(may be overkill, can wait until we have relevant changes)_

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,8 +109,12 @@ pip install fab-classic==1.17.0
 ```
 
 Create a `server_config.yaml` in the root of this repository using:
-`cp server_config_sample.yaml server_config.yaml`
-as a template
+```
+cp server_config_sample.yaml server_config.yaml
+```
+
+Below is an example `server_config.yaml` that defines 2 roles `comp-gpu` and `comp-cpu`,
+one with gpu style workers (`is_gpu` and the nvidia `docker_image`) and one with cpu style workers
 
 ```yaml
 comp-gpu:
@@ -129,7 +133,9 @@ comp-cpu:
   docker_image: codalab/competitions-v2-compute-worker:latest
 ```
 
-then you can execute commands against a group of servers:
+You can of course create your own `docker_image` and specify it here.
+
+You can execute commands against a role:
 
 ```bash
 ❯ fab -R comp-gpu status
--- a/README.md
+++ b/README.md
@@ -99,3 +99,44 @@ $ nvidia-docker run \
     --log-opt max-file=3 \
     codalab/competitions-v2-compute-worker:nvidia 
 ```
+
+# Worker management
+
+Outside of docker containers install [Fabric](http://fabfile.org/) like so:
+
+```bash
+pip install fab-classic
+```
+
+and create a `server_config.yaml` in the root of this respository:
+
+```yaml
+comp-gpu:
+  hosts:
+    - ubuntu@12.34.56.78
+    - ubuntu@12.34.56.79
+  broker_url: pyamqp://user:pass@host:port/vhost-gpu
+  is_gpu: true
+
+comp-cpu:
+  hosts:
+    - ubuntu@12.34.56.80
+  broker_url: pyamqp://user:pass@host:port/vhost-cpu
+  is_gpu: false
+```
+
+then you can execute commands against a group of servers:
+
+```bash
+❯ fab -R comp-gpu status
+..
+[ubuntu@12.34.56.78] out: CONTAINER ID        IMAGE                                           COMMAND                  CREATED             STATUS              PORTS               NAMES
+[ubuntu@12.34.56.78] out: 1d318268bee1        codalab/competitions-v2-compute-worker:nvidia   "/bin/sh -c 'celery …"   2 hours ago         Up 2 hours                              hardcore_greider
+..
+
+❯ fab -R comp-gpu update
+..
+(updates workers)
+```
+
+See available commands with `fab -l`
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -99,3 +99,46 @@ $ nvidia-docker run \
...
+
+# Worker management
+
+Outside of docker containers install [Fabric](http://fabfile.org/) like so:
+
+```bash
+pip install fab-classic
+```
+
+and create a `server_config.yaml` in the root of this respository:
```
jimmykodes: Add a `server_config_sample.yaml` file like we have `.env_sample` ?

fabfile.py
```diff
@@ -0,0 +1,68 @@
...
+#     docker_image: codalab/competitions-v2-compute-worker:nvidia
+#
+# You select the role to run with like s:
+#   $ fab -R role_name <command>
+env.roledefs = yaml.load(open('server_config.yaml').read())
+
+
+def status():
+    """Gets status of all docker containers on server"""
+    run("docker ps")
```
jimmykodes: this only gets status of running containers, not all containers. should either add `-a` to the command or update the docstring, depending on what is actually desired.

README.md
```diff
@@ -99,3 +99,46 @@ $ nvidia-docker run \
     --log-opt max-file=3 \
     codalab/competitions-v2-compute-worker:nvidia 
 ```
+
+# Worker management
+
+Outside of docker containers install [Fabric](http://fabfile.org/) like so:
+
+```bash
+pip install fab-classic
```
jimmykodes: Any way we can do this with a docker container instead of directly on our machines? at the very least, should probably specify a version number.

fabfile.py
```diff
@@ -0,0 +1,68 @@
...
+#   role_name:
+#     hosts:
+#       - ubuntu@12.34.56.78
+#     broker_url: pyamqp://user:pass@host:port/vhost-gpu
+#     is_gpu: true
+#     docker_image: codalab/competitions-v2-compute-worker:nvidia
+#
+# You select the role to run with like s:
+#   $ fab -R role_name <command>
+env.roledefs = yaml.load(open('server_config.yaml').read())
```
jimmykodes: do we need to specify a yaml loader here? (assuming we don't since it would have yelled at you, but just confirming.)

fabfile.py
```diff
@@ -0,0 +1,68 @@
...
+def update():
+    """Updates docker image on server and restarts the worker, based on server_config.yaml settings.
+
+    See README (Worker management section) for example settings."""
+    # Ensure that we're referencing one and only 1 role, so we don't accidentally update CPUs to GPUs
+    if not env.effective_roles:
+        print("ERROR: You must specify a role when running this task. I.e. fab -R some-gpu task_name`")
+        return
+    if len(env.effective_roles) > 1:
+        print("ERROR: Only specify 1 role (because we need 1 broker_url) when running this task")
```
jimmykodes: can we clean up some of the comments and verbiage in this doc to make it more clear that a "role" is either a cpu or gpu command? Role on its own is a little confusing.

README.md
ckcollab: added

README.md
ckcollab: good call, adding

fabfile.py
ckcollab: I didn't see any warnings (they could be squelched?) and it's user defined so I _think_ we should be ok, that's a good spot though

fabfile.py
ckcollab: added

fabfile.py
ckcollab: It's not either cpu/gpu though it could be any role you define from `server_config.yaml` -- each of those is a "role"

hmm..

README.md
```diff
@@ -99,3 +99,46 @@ $ nvidia-docker run \
...
+
+# Worker management
+
+Outside of docker containers install [Fabric](http://fabfile.org/) like so:
+
+```bash
+pip install fab-classic==1.17.0
+```
+
+and create a `server_config.yaml` in the root of this respository:
```
jimmykodes:
```suggestion
Create a `server_config.yaml` in the root of this repository using:
`cp server_config_sample.yaml server_config.yaml`
as a template
```

fabfile.py
jimmykodes: can leave it as is I suppose. I still think it is unintuitive, but it sounds like fabric is in general at the moment?

ckcollab: fabric v2 is unintuitive for sure, this is fabric v1 though which only took a few minutes to setup. I think roledefs isn't that wacky to understand, but maybe I'm just used to it. 

```yaml
role_name:
  hosts:
    - user@ip
    - user@ip2
```

```bash
# to run a task against all hosts within role_name
$ fab -R role_name task_name
```

Maybe I could define role better somewhere? will think about this for a sec

jimmykodes:
:shipit:

PR #364, opened 2020-02-27
Add current server time to competition detail page.

### LINKED ISSUE #268: Add current server time to comp detail page
In v1.5 of Codalab we put the current server time in competition detail pages, should do the same on V2:

image

brendon-lind:
@ckcollab 

Add current server time to competition detail page.

Closes #268 

# TODO
- [x] Add selenium test verifying the date is displayed on the detail page

- [ ] Refresh competition detail page

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [ ] Code review by reviewer
- [ ] Hand tested by reviewer
- [ ] Ready to merge

### DIFF
--- a/src/static/riot/competitions/detail/_header.tag
+++ b/src/static/riot/competitions/detail/_header.tag
@@ -42,7 +42,7 @@
                             </div>
                             <div>
                                 <span class="detail-label">Current server time:</span>
-                                <span class="detail-item">{pretty_date(CURRENT_DATE_TIME)}</span>
+                                <span class="detail-item" id="server_time">{pretty_date(CURRENT_DATE_TIME)}</span>
                             </div>
                             <div class="competition-secret-key" if="{ competition.admin }">
                                 <span class="secret-label">Secret url:</span>
--- a/src/static/riot/competitions/detail/_header.tag
+++ b/src/static/riot/competitions/detail/_header.tag
@@ -42,7 +42,7 @@
                             </div>
                             <div>
                                 <span class="detail-label">Current server time:</span>
-                                <span class="detail-item">{server_time}</span>
+                                <span class="detail-item">{pretty_date(CURRENT_DATE_TIME)}</span>
                             </div>
                             <div class="competition-secret-key" if="{ competition.admin }">
                                 <span class="secret-label">Secret url:</span>
@@ -168,12 +168,7 @@
         let self = this
 
         self.competition = {}
-        self.server_time = ''
         self.files = []
-
-        self.one("mount", function () {
-          self.get_current_server_time()
-        })
 
         CODALAB.events.on('competition_loaded', function (competition) {
             competition.admin = CODALAB.state.user.has_competition_admin_privileges(competition)
@@ -223,14 +218,6 @@
             document.execCommand("copy");
             window.getSelection().removeAllRanges();// to deselect
             $('.send-pop').popup('toggle')
-        }
-
-        self.get_current_server_time = function () {
-            CODALAB.api.get_current_server_time()
-                .done((data) => {
-                    self.server_time = data
-                    self.update()
-                })
         }
 
         self.get_end_date = function (competition) {
--- a/src/static/riot/competitions/detail/_header.tag
+++ b/src/static/riot/competitions/detail/_header.tag
@@ -40,6 +40,10 @@
                                 <span class="detail-label">Current phase ends:</span>
                                 <span class="detail-item">{get_end_date(competition)}</span>
                             </div>
+                            <div>
+                                <span class="detail-label">Current server time:</span>
+                                <span class="detail-item">{server_time}</span>
+                            </div>
                             <div class="competition-secret-key" if="{ competition.admin }">
                                 <span class="secret-label">Secret url:</span>
                                 <span id="secret-url">https://{ URLS.SECRET_KEY_URL(competition.id, competition.secret_key) }</span>
@@ -164,7 +168,12 @@
         let self = this
 
         self.competition = {}
+        self.server_time = ''
         self.files = []
+
+        self.one("mount", function () {
+          self.get_current_server_time()
+        })
 
         CODALAB.events.on('competition_loaded', function (competition) {
             competition.admin = CODALAB.state.user.has_competition_admin_privileges(competition)
@@ -214,6 +223,14 @@
             document.execCommand("copy");
             window.getSelection().removeAllRanges();// to deselect
             $('.send-pop').popup('toggle')
+        }
+
+        self.get_current_server_time = function () {
+            CODALAB.api.get_current_server_time()
+                .done((data) => {
+                    self.server_time = data
+                    self.update()
+                })
         }
 
         self.get_end_date = function (competition) {
--- a/src/templates/base.html
+++ b/src/templates/base.html
@@ -177,6 +177,7 @@
     <script>
         {# From `common_settings` context processor #}
         STORAGE_TYPE = "{{ STORAGE_TYPE }}"
+        CURRENT_DATE_TIME = "{% now "c" %}"
         CODALAB = {
             state: {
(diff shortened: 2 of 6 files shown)


src/static/riot/competitions/detail/_header.tag
```diff
@@ -216,6 +225,14 @@
             $('.send-pop').popup('toggle')
         }
 
+        self.get_current_server_time = function () {
+            CODALAB.api.get_current_server_time()
```
ckcollab: I think instead of doing a whole API call to get this, we can put current server time in `base.html` maybe? 

Can use the [now template tag.](https://docs.djangoproject.com/en/dev/ref/templates/builtins/?from=olddocs#now) 

```javascript
// I think this should work, "now()" shouldn't ever give us back a double quote
let CURRENT_DATE_TIME = "{% now "jS F Y H:i" %}"
```

Nice job implementing this pattern, though!

ckcollab:
image

_guffaw_ :)

src/static/riot/competitions/detail/_header.tag
ckcollab: It might even be nicer to return some isoformat date that way we can have a more robust Date object instead of a string. I forget if we have arrow/moment/whatever JS datetime lib..

ckcollab:
Also, for practice, you could make sure in the selenium tests that this time appears on the page properly. added a todo to the main PR description for you

src/static/riot/competitions/detail/_header.tag
```diff
@@ -40,6 +40,10 @@
                                 <span class="detail-label">Current phase ends:</span>
                                 <span class="detail-item">{get_end_date(competition)}</span>
                             </div>
+                            <div>
+                                <span class="detail-label">Current server time:</span>
+                                <span class="detail-item" id="server_time">{pretty_date(CURRENT_DATE_TIME)}</span>
```
brendon-lind: This `id` is here for selenium tests.

ckcollab: cool

src/tests/functional/test_competitions.py
```diff
@@ -12,6 +13,15 @@ def setUp(self):
...
         self.login(self.user.username, 'test')
 
+    def current_server_time_exists(self):
+        # Get server time element
+        element = self.find('#server_time')
+        text = element.get_attribute('innerText')
+
+        # Check that the text is a valid datetime by loading it with strptime.
+        # This will raise a ValueError if the format is incorrect.
+        datetime.strptime(text, '%B %d, %Y, %I:%M %p %Z')
```
ckcollab: could you do an assertion of some kind here? at least:

```datetime.strptime(text, '%B %d, %Y, %I:%M %p %Z')```

File: src/static/riot/competitions/detail/_header.tag (hunk 2 above)
Comment (ckcollab) [inline src/static/riot/competitions/detail/_header.tag]: where did `pretty_date` come from? cool

File: src/static/riot/competitions/detail/_header.tag (hunk 1 above)
Comment (gibsonbailey) [reply src/static/riot/competitions/detail/_header.tag]: ✅

File: src/tests/functional/test_competitions.py (hunk 3 above)
Comment (gibsonbailey) [reply src/tests/functional/test_competitions.py]: Was this comment left mistakenly unfinished?

File: src/static/riot/competitions/detail/_header.tag (hunk 1 above)
Comment (gibsonbailey) [reply src/static/riot/competitions/detail/_header.tag]: @brendon-lind @dde6khkg After addressing a comment, at the very least, we place a checkmark (***after implementing the change***) like the one above, just so the reviewer knows that it has been addressed.

File: src/tests/functional/test_competitions.py (hunk 3 above)
Comment (ckcollab) [reply src/tests/functional/test_competitions.py]: I left out `assert` in my example :( oops

File: src/static/riot/competitions/detail/_header.tag (hunk 1 above)
Comment (brendon-lind) [reply src/static/riot/competitions/detail/_header.tag]: > It might even be nicer to return some isoformat date that way we can have a more robust Date object instead of a string. I forget if we have arrow/moment/whatever JS datetime lib..

Done

Comment (brendon-lind) [reply src/static/riot/competitions/detail/_header.tag]: Done!

File: src/static/riot/competitions/detail/_header.tag (hunk 2 above)
Comment (brendon-lind) [reply src/static/riot/competitions/detail/_header.tag]: This was a already created function in `ours/utils.js`

File: src/tests/functional/test_competitions.py (hunk 3 above)
Comment (brendon-lind) [reply src/tests/functional/test_competitions.py]: Done. Repassed selenium tests.

**Comment by ckcollab** [comment]
I'm on a new box doing an a days worth of installs right now so I can't build this locally.. could you put a screenshot of how the date is placed on the page in a normal desktop resolution, tablet resolution, and mobile resolution? If not I can pull this local sometime tomorrow probably and take a looksy :)

**Comment by brendon-lind** [comment]
Desktop
ipad
S5android

Looks like the scaling on the whole site for mobile and tablet are not scaling correctly.

File: src/tests/functional/test_competitions.py [hunk 4]
```diff
@@ -12,6 +13,15 @@ def setUp(self):
...
         self.login(self.user.username, 'test')
 
+    def current_server_time_exists(self):
+        # Get server time element
+        element = self.find('#server_time')
+        text = element.get_attribute('innerText')
+
+        # Check that the text is a valid datetime by loading it with strptime.
+        # This will raise a ValueError if the format is incorrect.
+        assert datetime.strptime(text, '%B %d, %Y, %I:%M %p %Z')
```
jimmykodes: The problem with this test is that if it is in a fail state, we don't get the nice pytest interpreted assertion error, we just get a value error, as the comment notes. we want to avoid that, since this type of a failure is slower to locate and fix. my preferred solution would be something like the following pseudo-code:
```python
try:
    # assert that the time displayed _is_ server time. possible problem here: I believe pretty date 
    # shifts timezones to the user's local timezone, maybe convert to unix timestamps to compare.
    assert strptime(...) == timezone.now()
except ValueError:
    assert False, "Date not displayed in the correct format"
```
This tests that the server time is present, correct, _and_ in the proper format.

src/tests/functional/test_competitions.py
```diff
@@ -27,6 +37,7 @@ def _upload_competition(self, competition_zip_path):
         comp_url = reverse("competitions:detail", kwargs={"pk": comp.id})
         self.find(f'a[href="{comp_url}"]').click()
         self.assert_current_url(comp_url)
+        self.current_server_time_exists()
```
jimmykodes: this test should be run as a standalone, not a part of the upload competitions step. If this test is failing, we should not have to re-run this _very_ long running test.

things that should be tested would be: 
- server time visible if anonymous user
- server time visible if logged in

src/templates/base.html
```diff
@@ -177,6 +177,7 @@ <h4 class="ui inverted header">Codalab v2</h4>
     <script>
         {# From `common_settings` context processor #}
         STORAGE_TYPE = "{{ STORAGE_TYPE }}"
+        CURRENT_DATE_TIME = "{% now "c" %}"
```
jimmykodes: @ckcollab Do we want/need this accessible from every single page? or just the comp detail page? could maybe pass this in opts on the html page w/ the comp-detail riot tag?

src/tests/functional/test_competitions.py
gibsonbailey: I'll check if we can gain access to a js variable in selenium, so we can make failure states of this test more informative. It will go something like the following:

1. Check for `CURRENT_SERVER_TIME` variable in js directly.
2. Check that there is content in the `#server_time` element (`len(text) > 0`).
3. Check that `text` is formatted properly as a datestring by loading it with `strptime`.

src/templates/base.html
ckcollab: I think it could be useful, yeah, what do you think?

PR #367, opened 2020-02-28
Execution Time Limit Fixes

### LINKED ISSUE #366: Fix references to "execution_time_limit_ms"
the execution time limit is seconds, not microseconds, so remove the 'ms' from references.

### LINKED ISSUE #374: Remove redundant/unused buttons on user dropdown
Queue management button is redundant
Switch User is unused.

Screen Shot 2020-03-03 at 8 26 26 AM

jimmykodes:
# @ mention of reviewers
@ckcollab 

Fixes references to execution time limit as _ms since the actual unit is seconds.
Should address uploading a comp, and creating a dump.
Also actually times out ingestion/scoring programs and writes their logs.

Also removes redundant buttons in user dropdown. Not strictly related to this PR, but since it is small and needed to be done, addressing it here rather than creating a new PR.

# Issues this PR resolves
Resolves #366 
wrong number in the branch name... oops.
resolves #374 

- [x] upload a competition, and set its phase's execution time limit to something like 10 seconds.
- [x] upload a submission with a sleep of that length contained in it.
- [x] submission should terminate and the docker container should be removed.

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/docker/compute_worker/compute_worker.py
+++ b/docker/compute_worker/compute_worker.py
@@ -615,13 +615,13 @@ class Run:
                     elapsed_time = self.execution_time_limit
                 return_code = logs["proc"].returncode
                 if return_code is None:
-                    # procedure is still running, kill it
                     logger.info('No return code from Process. Killing it')
                     if kind == 'ingestion':
                         program_to_kill = self.ingestion_container_name
                     else:
                         program_to_kill = self.program_container_name
-                    kill_code = subprocess.call(['docker', 'kill', str(program_to_kill)])
+                    # Try and stop the program. If stop does not succeed
+                    kill_code = subprocess.call(['docker', 'stop', str(program_to_kill)])
                     logger.info(f'Kill process returned {kill_code}')
                 if kind == 'program':
                     self.program_exit_code = return_code
--- a/docker/compute_worker/compute_worker.py
+++ b/docker/compute_worker/compute_worker.py
@@ -5,6 +5,7 @@ import json
 import logging
 import os
 import shutil
+import signal
 import socket
 import tempfile
 import time
@@ -127,6 +128,14 @@ def delete_files_in_folder(folder):
             shutil.rmtree(file_path)
 
 
+class ExecutionTimeLimitExceeded(Exception):
+    pass
+
+
+def alarm_handler(signum, frame):
+    raise ExecutionTimeLimitExceeded
+
+
 class Run:
     """A "Run" in Codalab is composed of some program, some data to work with, and some signed URLs to upload results
     to. There is also a secret key to do special commands for just this submission.
@@ -149,6 +158,7 @@ class Run:
         self.root_dir = tempfile.mkdtemp(dir="/tmp/codalab-v2")
         self.input_dir = os.path.join(self.root_dir, "input")
         self.output_dir = os.path.join(self.root_dir, "output")
+        self.logs = {}
 
         # Details for submission
         self.is_scoring = run_args["is_scoring"]
@@ -323,7 +333,7 @@ class Run:
             stderr=asyncio.subprocess.PIPE
         )
 
-        logs = {
+        self.logs[kind] = {
             "stdout": {
                 "data": b'',
                 "stream": proc.stdout,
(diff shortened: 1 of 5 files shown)


docker/compute_worker/compute_worker.py
```diff
@@ -607,7 +616,20 @@ def start(self):
             self.watch_file(),
             loop=loop,
         )
-        loop.run_until_complete(gathered_tasks)
+        signal.signal(signal.SIGALRM, alarm_handler)
+        signal.alarm(self.execution_time_limit)
+        try:
+            loop.run_until_complete(gathered_tasks)
+        except ExecutionTimeLimitExceeded:
```
ckcollab: May make sense to do file uploading here instead of on exception. Like put it in a finally clause? That way we don't have to set the key to None as well

docker/compute_worker/compute_worker.py
```diff
@@ -127,6 +128,14 @@ def delete_files_in_folder(folder):
             shutil.rmtree(file_path)
 
 
+class ExecutionTimeLimitExceeded(Exception):
+    pass
+
+
+def alarm_handler(signum, frame):
```
ckcollab: I assume the alarm stuff doesn't take a lambda for some reason? Or did python get rid of inline lambda, can't recall now..

jimmykodes: tried to do a lambda, it didn't like it. At all...

docker/compute_worker/compute_worker.py
```diff
@@ -342,9 +357,12 @@ def _get_bundle(self, url, destination, cache=True):
         websocket = await websockets.connect(self.websocket_url)
         websocket_errors = (socket.gaierror, websockets.WebSocketException, websockets.ConnectionClosedError, ConnectionRefusedError)
 
-        while any(v["continue"] for v in logs.values()):
+        while any(v["continue"] for k, v in self.logs[kind].items() if k in ['stdout', 'stderr']):
             try:
-                for value in logs.values():
+                for key, value in self.logs[kind].items():
+                    # todo: just grab the keys we care about instead of "continuing" from keys we don't care about
```
ckcollab: Is this meaning to do something like:

```python
logs = [self.logs[kind][key] for key in ('stdout', 'stderr')]
for value in logs:
```

hope that's helpful... if not should we resolve this todo in this PR or push it off?

docker/compute_worker/compute_worker.py
```diff
@@ -607,7 +602,41 @@ def start(self):
...
+        signal.signal(signal.SIGALRM, alarm_handler)
+        signal.alarm(self.execution_time_limit)
+        try:
+            loop.run_until_complete(gathered_tasks)
+        except ExecutionTimeLimitExceeded:
+            pass
+        finally:
+            self.watch = False
+            for kind, logs in self.logs.items():
+                elapsed_time = logs["end"] - logs["start"] if logs["end"] is not None else self.execution_time_limit
```
ckcollab: i don't think this needs to be an inline if, could be easier to understand broken out? what do you think?

docker/compute_worker/compute_worker.py
```diff
@@ -607,7 +602,41 @@ def start(self):
...
+                return_code = logs["proc"].returncode
+                if return_code is None:
+                    # procedure is still running, kill it
+                    logger.info('No return code from Process. Killing it')
+                    kill_code = subprocess.call(['docker', 'kill', str(self.ingestion_container_name if kind == "ingestion" else self.program_container_name)])
```
ckcollab: would prefer something like...

```python
if kind == 'ingestion':
    program_to_kill = self.ingestion_container_name
else:
    program_to_kill = self.program_container_name
```

docker/compute_worker/compute_worker.py
```diff
@@ -607,7 +602,41 @@ def start(self):
             self.watch_file(),
             loop=loop,
         )
-        loop.run_until_complete(gathered_tasks)
+        signal.signal(signal.SIGALRM, alarm_handler)
+        signal.alarm(self.execution_time_limit)
+        try:
+            loop.run_until_complete(gathered_tasks)
+        except ExecutionTimeLimitExceeded:
+            pass
```
ckcollab: instead of passing here we should let the user know why it failed i suppose?

```python
self._update_status(STATUS_FAILED, f"Time limit exceeded. Limit was {self.execution_time_limit}ms.")
```

this might get overwritten somewhere else in the script though?

docker/compute_worker/compute_worker.py
jimmykodes: Yea, that was what I meant to go back and add. Completely missed this Todo. Good catch. Will add.

docker/compute_worker/compute_worker.py
jimmykodes: `raise SubmissionException(f"Execution Time Limit exceeded. Limit was {self.execution_time_limit} seconds")` as a solution instead, since we catch a Submission Exception to call `_update_status(STATUS_FAILED)`

ckcollab:
:shipit: 
working locally

PR #412, opened 2020-06-02
[Feature]: Support transform a benchmark bundle to competition.

AuroraLauu:
Hi there, I am developer from 4Paradigm for Codabench project, can you guys have a code review for my commit? many thanks!!!

### DIFF
--- a/src/apps/competitions/unpackers/codabench_v1.py
+++ b/src/apps/competitions/unpackers/codabench_v1.py
@@ -1,32 +1,14 @@
-import os
 
 from django.utils import timezone
 
-from competitions.unpackers.base_unpacker import BaseUnpacker
+from competitions.unpackers.v2 import V2Unpacker
 from .utils import CompetitionUnpackingException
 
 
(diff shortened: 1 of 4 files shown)


src/apps/competitions/unpackers/codabench_v1.py
```diff
@@ -0,0 +1,211 @@
+import os
+
+from django.utils import timezone
+
+from competitions.unpackers.base_unpacker import BaseUnpacker
+from .utils import CompetitionUnpackingException
+
+
+class CodabenchV1Unpacker(BaseUnpacker):
```
ckcollab, line 9: this seems to have quite a few DRY violations?

AuroraLauu: Hi Eric, From your opinion,  a better approach would be to inherit V2Unpacker and then add the features I need on top of that, right?

AuroraLauu: Hi Eric, i make a changed already, can u review it, thanks.

PR #475, opened 2020-07-17
detailed results fixes

### LINKED ISSUE #466: detailed results from v1.5 handled by compute_worker watch_file()
Detailed results for v1.5 competitions does not output to a `detailed_results.html`, they instead output to some arbitrarily named `*.html` file. Codalab should be able to handle both formats.

dde6khkg:
@ mention of reviewers
@ckcollab 

This PR fixes detailed results output for v1.5 competitions.

# Issues this PR resolves
#466 

- [x] Upload a v1.5 competition with detailed results enabled
- [x] Submit to the competition
- [x] Make sure detailed results output is visible under the competition output/in the submission modal under `VISUALIZATION`

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge


docker/compute_worker/compute_worker.py
```diff
@@ -205,22 +206,31 @@ def __init__(self, run_args):
...
         if not self.detailed_results_url:
             return
         file_path = os.path.join(self.output_dir, 'detailed_results.html')
+        v15_file_path = None
         last_modified_time = None
         start = time.time()
         expiration_seconds = 60
+
         while self.watch:
+            if self.completed_program_counter >= 2:
```
ckcollab, line 221: should prob just add this to the while?

```
while self.watch and self.completed_program_counter < 2:
```

docker/compute_worker/compute_worker.py
```diff
@@ -205,22 +206,31 @@ def __init__(self, run_args):
...
 
-    # TODO: Rename watch_file to "detailed_results_watcher" or some such
-    async def watch_file(self):
+    async def watch_detailed_results(self):
         """Watches files alongside scoring + program docker containers, currently only used
         for detailed_results.html"""
         if not self.detailed_results_url:
             return
         file_path = os.path.join(self.output_dir, 'detailed_results.html')
+        v15_file_path = None
```
ckcollab, line 215: I think we don't need this extra variable? Have `file_path` be the normal default, if nothing found at this path re-assign file_path to the first HTML file found? maybe add comments about this weird behavior

docker/compute_worker/compute_worker.py
```diff
@@ -230,7 +240,10 @@ async def watch_file(self):
             await asyncio.sleep(5)
         else:
             # make sure we always send the final version of the file
-            await self.send_detailed_results(file_path)
+            if os.path.exists(file_path):
+                await self.send_detailed_results(file_path)
+            elif os.path.exists(v15_file_path):
```
ckcollab, line 245: again with above change we won't need 2 cases here

docker/compute_worker/compute_worker.py
```diff
@@ -407,12 +420,20 @@ async def _run_docker_cmd(self, docker_cmd, kind):
 
         logger.info(f"Process exited with {proc.returncode}")
         logger.info(f"Disconnecting from websocket {self.websocket_url}")
+
+        # Communicate that the program is closing
+        self.completed_program_counter += 1
```
ckcollab, line 425: nice comment, ty!

docker/compute_worker/compute_worker.py
```diff
@@ -215,22 +215,35 @@ async def watch_file(self):
         last_modified_time = None
         start = time.time()
         expiration_seconds = 60
-        while self.watch:
+
+        last_run = False
```
ckcollab, line 219: all this `last_run` logic _feels weird_, hmm...

If the point is to allow programs to clean up and send the "final detailed results" can we have a sleep of 5-10 seconds at the end?

dde6khkg: The point of the `last_run` flag is to make sure the while block gets run at least once after the scoring program finishes. I was running into a case where the scoring program would create the `*.html` file and finish during the 5 second sleep at the end of the while loop, so the `file_path` would never get updated to the v1.5 style. Is there a problem with doing it this way? Maybe we could run a different loop that waits for the file to appear, then it enters this loop?

ckcollab: I see, well we could wait 10 seconds after the loop ends to get basically the same behavior?

dde6khkg: The problem is the path is updated inside the loop block. So the `file_path` doesn't exist, therefore it doesn't enter that `glob.glob` block. Then, the loop will exit, wait 10 seconds, and still not have the correct `file_path`. So we need some way to intelligently wait for one of these files to exist before sending detailed results and exiting the function.

PR #478, opened 2020-07-21
refactor tasks & datasets

### LINKED ISSUE #428: Refactoring tasks + datasets
- [x] "Tasks & Datasets" tab should be renamed to "Resources"
- [ ] Tasks require "unit_test" / "solution" and a compute worker run is automatically ran when created

It should not be possible to:
- [x] Edit or delete a file used by a task or an algorithm
- [x] Delete a task used in a competition
- [x] Delete an algorithm that was submitted to a competition
- [x] Publish a task not validated by an algorithm
- [x] Publish an algorithm not validated by a task
- [x] It should be possible to delete multiple files at the same time.

dde6khkg:
# @ mention of reviewers
@ckcollab 

This PR adds some limitations when deleting things from the "Tasks and Datasets" page. It also allows the deletion of multiple items at once.

# Issues this PR resolves
#428 

# Suggested changes
I find it weird that a submission shouldn't be deletable after it's been assigned to a competition. Users aren't going to be able to delete any submissions until the competition is deleted. It makes more sense to me to disallow deletions of submissions that have been added to the leaderboard, or something along those lines.

- [x] Try to delete some tasks, datasets, and files
- [x] Make sure a file can't be deleted if it's used by a task or algorithm
- [x] Make sure a task can't be published if it hasn't been validated
- [x] Make sure a task can't be deleted if it's used in a competition
- [x] Make sure a submission can't be published if it's not validated
- [x] Make sure a submission can't be deleted if it's part of a competition
- [x] Make sure multiple tasks and submissions can be deleted at once

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by ~reviewer~ Logan
- [x] Ready to merge


src/apps/api/serializers/tasks.py
```diff
@@ -65,6 +66,12 @@ class Meta:
             'created_by',
         )
 
+    def validate_is_public(self, is_public):
+        validated = Task.objects.get(id=self.instance.id)._validated
+        if is_public and not validated:
+            raise ValidationError('Task must be validated before it can be published')
```
ckcollab, line 72: nice

src/apps/api/views/datasets.py
```diff
@@ -63,10 +62,48 @@ def create(self, request, *args, **kwargs):
...
+            qs.delete()
+
+        return Response(
+            errors if errors else {'detail': 'Datasets deleted successfully'},
+            status=status.HTTP_400_BAD_REQUEST if errors else status.HTTP_200_OK
+        )
+
+    def check_delete_permissions(self, request, dataset):
+        if request.user != dataset.created_by:
+            return 'Cannot delete a dataset that is not yours'
```
ckcollab, line 97: I think we can simplify this a bit by raising a `ValidationError(msg)` or some such? `PermissionDenied(msg)` maybe?

src/apps/api/views/tasks.py
```diff
@@ -66,6 +69,37 @@ def update(self, request, *args, **kwargs):
...
+            qs.delete()
+
+        return Response(
+            errors if errors else {'detail': 'Tasks deleted successfully'},
+            status=status.HTTP_400_BAD_REQUEST if errors else status.HTTP_200_OK
+        )
+
+    def check_delete_permissions(self, request, task):
+        if request.user != task.created_by:
+            return "Cannot delete a task that is not yours"
```
ckcollab, line 103: same `ValidationError` comment as above

src/static/riot/datasets/management.tag
```diff
@@ -19,6 +19,10 @@
         <i selenium="add-dataset" class="plus icon"></i>
         Add Dataset
     </button>
+    <button class="ui red right floated labeled icon button {disabled: marked_datasets.length === 0}" onclick="{delete_datasets}">
```
ckcollab, line 22: awesome, nice job disabling/enabling based on selection!

src/static/riot/datasets/management.tag
```diff
@@ -52,6 +57,12 @@
                     <i class="icon delete"></i>
                 </button>
             </td>
+            <td class="center aligned">
+                <div class="ui fitted checkbox">
+                    <input type="checkbox" name="delete_checkbox" onclick="{ mark_dataset_for_deletion.bind(this, dataset) }">
+                    <label></label>
```
ckcollab, line 63: label needed?

src/static/riot/datasets/management.tag
```diff
@@ -214,7 +226,10 @@
             self.update_datasets()
         })
 
-        self.show_info_modal = function (row) {
+        self.show_info_modal = function (row, e) {
+            if (e.target.type === 'checkbox') {
+                return
```
ckcollab, line 231: nice! Can you add a comment about why we did this? although it is probably pretty clear, may not be in 2 years

src/static/riot/tasks/management.tag
```diff
@@ -31,6 +37,12 @@
                     <i class="icon delete"></i>
                 </button>
             </td>
+            <td class="center aligned">
+                <div class="ui fitted checkbox">
+                    <input type="checkbox" name="delete_checkbox" onclick="{ mark_task_for_deletion.bind(this, task) }">
+                    <label></label>
```
ckcollab, line 43: pull?

src/apps/api/views/datasets.py
dde6khkg: I figured it should be done this way so we can return an object with every failure. I believe if we raise an error, it won't continue through the rest of the loop, and we only get one failure. Is that an okay way to do this?

ckcollab: Ah, right, good point, makes sense. Maybe add a comment about this?

src/static/riot/datasets/management.tag
dde6khkg: I believe the checkboxes disappear if the label tag is deleted. Like this: 
image

Logan-Ruf:
Pressing the "delete box" instead of selecting the Dataset then pressing "Delete Selected Datasets" does not return the reason why the dataset can't be deleted.
Screen Shot 2020-07-30 at 8 26 51 PM
Screen Shot 2020-07-30 at 8 26 43 PM

Logan-Ruf:
Probably should have an error message for when you try to make a task public that isn't validated.
Screen Shot 2020-07-30 at 9 08 41 PM

PR #614, opened 2020-12-22
Submission status cleanup

### LINKED ISSUE #608: Submission status cleanup
Each hour execute a job that sets submission statuses = Failed for submissions that have been running for .. 24 hours after their possible max run time?

dde6khkg:
# @ mention of reviewers
@Logan-Ruf @ckcollab 

This PR adds a celery beat task to set submission statuses to FAILED if the submission has been running for 24 hours past the max run time.

# Issues this PR resolves
#608

- [x] Run `test_submissions_are_cancelled_if_running_24_hours_past_execution_time_limit` and `test_cancelling_parent_submission_cancels_all_children`

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge


src/apps/competitions/tests/test_submissions.py
```diff
@@ -380,3 +382,14 @@ def test_fact_sheet_with_blank_required_text_is_not_valid(self):
         }
         serializer = SubmissionCreationSerializer(data=submission, instance=Submission)
         assert not serializer.is_valid()
+
+
+class TestSubmissionTasks(SubmissionTestCase):
+    def test_submissions_are_cancelled_if_running_24_hours_past_execution_time_limit(self):
+        self.submission = self.make_submission()
+        self.submission.status = Submission.RUNNING
+        self.submission.created_when = timezone.now() - timedelta(milliseconds=(3600000 * 24) + self.submission.phase.execution_time_limit)
```
ckcollab, line 391: can you modify the test so there is a running submission that is ignored?

dde6khkg: Ah yeah, good idea.

src/apps/competitions/tests/test_submissions.py
```diff
@@ -11,6 +11,8 @@
 from api.serializers.competitions import CompetitionSerializer
 from competitions.models import Submission
 from competitions.tasks import run_submission
+
+from apps.competitions.tasks import submission_status_cleanup
```
Logan-Ruf, line 15: Can this be one line?
```suggestion
from competitions.tasks import run_submission, submission_status_cleanup
```

Logan-Ruf (changes requested):

src/apps/competitions/tasks.py
```diff
@@ -744,3 +745,15 @@ def update_phase_statuses():
...
         comp.update_phase_statuses()
+
+
+@app.task(queue='site-worker')
+def submission_status_cleanup():
+    submissions = Submission.objects.filter(status=Submission.RUNNING).select_related('phase')
+
+    for sub in submissions:
+        # Check if the submission has been running for 24 hours longer than execution_time_limit
+        if sub.created_when < now() - timedelta(milliseconds=(3600000 * 24) + sub.phase.execution_time_limit):
```
ckcollab, line 756: hmm, I wonder if `created_when` works properly here, what if a submission was created, queued up for a day or two, then started? do we need a `started_when`?

Logan-Ruf (approved):

src/apps/competitions/tasks.py
dde6khkg: Yeah that could definitely happen. Adding a `started_when`

src/apps/competitions/tasks.py
```diff
@@ -744,3 +745,20 @@ def update_phase_statuses():
     competitions = Competition.objects.exclude(phases__in=Phase.objects.filter(is_final_phase=True, end__lt=now()))
     for comp in competitions:
         comp.update_phase_statuses()
+
+
+@app.task(queue='site-worker')
+def submission_status_cleanup():
+    submissions = Submission.objects.filter(status=Submission.RUNNING).exclude(has_children=True).select_related('phase')
```
ckcollab, line 752: nitpick, we could do it like this in the future :D (remove `exclude()`) -- I also added `select_related` for `parent`
```suggestion
    submissions = Submission.objects.filter(status=Submission.RUNNING, has_children=False).select_related('phase', 'parent')
```

src/apps/competitions/tasks.py
```diff
@@ -744,3 +745,20 @@ def update_phase_statuses():
...
+
+    for sub in submissions:
+        # Check if the submission has been running for 24 hours longer than execution_time_limit
+        if sub.started_when < now() - timedelta(milliseconds=(3600000 * 24) + sub.phase.execution_time_limit):
+            if sub.parent is not None:
+                app.control.revoke(sub.parent.celery_task_id, terminate=True)
+                sub.parent.status = Submission.FAILED
+                sub.parent.save()
+            else:
```
ckcollab, line 761: don't we want to run this side of the else in both cases?

ckcollab: ..and the DRY violations here make me think we should have a method on `Submission` model that does this..

```python
class Submission...
    def fail_and_cancel_task(self):
            app.control.revoke(self.celery_task_id, terminate=True)
            self.status = Submission.FAILED
            self.save()
```

src/apps/competitions/migrations/0028_submission_started_when.py
```diff
@@ -0,0 +1,18 @@
...
+    dependencies = [
+        ('competitions', '0027_auto_20201219_1308'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='submission',
+            name='started_when',
+            field=models.DateTimeField(null=True),
+        ),
```
Logan-Ruf, line 17: Develop has another 0028 migration now. Merge and remake this

Logan-Ruf (changes requested):

src/apps/competitions/tasks.py
dde6khkg: Ah yeah I saw someone use that when I was looking for exclude, but it sounds like they do the same thing and one doesn't necessarily perform better. I do like the way that looks better though so I'm down.

src/apps/competitions/tasks.py
dde6khkg: I was assuming running this on the parent cancelled the child because I believe we use the `Submission.cancel()` method on only the parent submission when cancelling submissions run on multiple tasks.

src/apps/competitions/models.py
```diff
@@ -511,11 +512,10 @@ def re_run(self, task=None):
...
-    def cancel(self):
-        from celery_config import app
+    def cancel(self, status=CANCELLED):
         if self.status not in [Submission.CANCELLED, Submission.FAILED, Submission.FINISHED]:
             app.control.revoke(self.celery_task_id, terminate=True)
-            self.status = Submission.CANCELLED
+            self.status = status
             self.save()
             return True
         return False
```
Logan-Ruf, line 522: Does this affect the children as well? If the parent submission is set to fail it would make sense to fail and stop all the children.

Logan-Ruf (changes requested):

Logan-Ruf (approved):
Looks good!

PR #619, opened 2020-12-29
Changed styling for forum email

### LINKED ISSUE #583: Forum email improvements
image

Looks like the post body is missing from the emails?

Also, we should probably fix up the fonts/consistency, could maybe be a fun opportunity to implement this:
https://grtcalculator.com/

Logan-Ruf:
# @ mention of reviewers
@ckcollab 

# Issues this PR resolves
Resolves #583 

image

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [ ] Code review by reviewer
- [ ] Hand tested by reviewer
- [ ] Ready to merge

### DIFF
--- a/src/static/riot/competitions/editor/_competition_details.tag
+++ b/src/static/riot/competitions/editor/_competition_details.tag
@@ -320,6 +320,7 @@
                 }
             }
             self.update()
+            self.form_updated()
             // Put dropdown 'set selected' here so it doesn't trigger on_change:form_updated() before fact_sheet_questions is set
             $(self.refs.competition_type).dropdown('set selected', competition.competition_type)
             // Form change events
--- a/src/templates/emails/base_email.html
+++ b/src/templates/emails/base_email.html
@@ -17,14 +17,20 @@
           palette teal: #67B8C7
         */
         html, body {
+            font-size: 18px;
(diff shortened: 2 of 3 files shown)


src/static/riot/competitions/editor/_competition_details.tag
```diff
@@ -320,6 +320,7 @@
                 }
             }
             self.update()
+            self.form_updated()
```
ckcollab, line 323: we need this back ?

ckcollab:
I think make "Codalab Team" into "Codabench Team" (and remove bold) and then :shipit:

src/static/riot/competitions/editor/_competition_details.tag
Logan-Ruf: Yeah, the dropdown has a default value. So it only calls `forum_updated()` when the competition doesn't have that default value. So if we have it there it will call `form_updated` 1 or 2 times, instead of 0 or 1 times. If it isn't called the form bugs out.

ckcollab: ah should push this directly to develop then?

PR #631, opened 2021-01-11
Submission Logs now save to redis

### LINKED ISSUE #492: Move submission log storage from temp dir to Redis
### Overview

To improve end user experience and not drop logs between deployments, we should store logs in Redis.

### Relevant sections of code:

This websocket endpoint takes a list of submission IDs and returns their data from temp file storage. This needs to be changed to use redis instead of temp file storage.

```python
    async def receive(self, text_data=None, bytes_data=None):
        """We expect to receive a message at this endpoint containing the ID(s) of submissions to get
        details about; typically on page load, looking up the previous submission details"""
        data = json.loads(text_data)

        submission_ids = data.get("submission_ids", [])

        if submission_ids:
            # Filter out submissions not by this user
            submissions = Submission.objects.filter(id__in=submission_ids, owner=self.scope["user"])

            for sub in submissions:
                text_path = os.path.join(settings.TEMP_SUBMISSION_STORAGE, f"{sub.id}.txt")
                if os.path.exists(text_path):
                    with open(text_path) as f:
                        text = f.read()
                    await self.group_send(text, sub.id, full_text=True)
```

### Tasks

- [ ] Make websocket endpoint _read_ from redis
- [ ] Make websocket endpoint _write_ to redis
- [ ] Ensure redis on Heroku is configured to dump the oldest data when we go over data limits

Logan-Ruf:
# @ mention of reviewers
@ckcollab 

Submission logs now save to redis

# Issues this PR resolves
Resolves #492 

- [ ] add checklist here

# Checklist
- [ ] Code review by me 
- [ ] Hand tested by me 
- [ ] I'm proud of my work
- [ ] Code review by reviewer
- [ ] Hand tested by reviewer
- [ ] Ready to merge


src/apps/competitions/consumers.py
```diff
@@ -39,16 +37,13 @@ async def receive(self, text_data=None, bytes_data=None):
...
-        os.makedirs(os.path.dirname(submission_output_path), exist_ok=True)
         data = json.loads(text_data)
         if data['kind'] == 'detailed_result_update':
             data['result_url'] = make_url_sassy(Submission.objects.get(id=submission_id).detailed_result.name)
             # update text data to include the newly added sas url for retrieval on page refresh
             text_data = json.dumps(data)
 
-        async with aiofiles.open(submission_output_path, 'a+') as f:
-            await f.write(f'{text_data}\n')
+        cache.set(f'submission-{submission_id}-log', f'{text_data}\n', 3600 * 25)
```
ckcollab, line 46: what this replaces was appending to a file I believe, can we append to the cache?

Painfully, after looking at the django redis libs, we may have to switch to jazzband to get raw redis client support?
https://github.com/jazzband/django-redis#raw-client-access

we should set it up to also nicely use connection pools like how we have the current library:
https://github.com/jazzband/django-redis#connection-pools

aaaand probably compress data:
https://github.com/jazzband/django-redis#compression-support

maybe it'd make sense to store this data as a list of lines and RPUSH?
https://stackoverflow.com/a/41374478/1359986

ckcollab (changes requested):

src/apps/competitions/consumers.py
```diff
@@ -87,12 +82,12 @@ async def receive(self, text_data=None, bytes_data=None):
...
             # Filter out submissions not by this user
             submissions = Submission.objects.filter(id__in=submission_ids, owner=self.scope["user"])
+            con = get_redis_connection("default")
 
             for sub in submissions:
-                text_path = os.path.join(settings.TEMP_SUBMISSION_STORAGE, f"{sub.id}.txt")
-                if os.path.exists(text_path):
-                    with open(text_path) as f:
-                        text = f.read()
+                text = (con.get(f':1:submission-{sub.id}-log'))
```
ckcollab, line 88: can use `cache.get` here?

src/apps/competitions/consumers.py
```diff
@@ -87,12 +82,12 @@ async def receive(self, text_data=None, bytes_data=None):
...
+                if text:
+                    text = str(text, 'utf-8')
                     await self.group_send(text, sub.id, full_text=True)
```
ckcollab, line 91: I think you can inline do `text.encode('utf-8')` or whatever?

```py
await self.group_send(text.encode('utf-8'), sub.id, full_text=True)
```

Logan-Ruf: If text returns `None` then that will throw an error.

src/apps/competitions/consumers.py
Logan-Ruf: For some reason `cache.get` tries to decode the string and fails.

Issue #690, opened 2022-03-28
Unable to create compute workers


Didayolo:
I have trouble creating compute workers.

* For the default queue, I struggle to find the broker URL
* For a custom queue, I have a broker URL of this form:

```
pyamqp://26df2d3e-345e-472b-[...]-6e1ae58c4202:6a7f18d6-85a0-455f-[...]-9903297abd15@rabbit:5672/f561486f-803a-[...]-ab79-d8fecb6badf1
```

Shouldn't the domain appears somewhere here? Like `codabench.org` instead of `rabbit`?

I get the following error:

```
[2022-03-28 14:38:15,413: ERROR/MainProcess] consumer: Cannot connect to amqp://26df2d3e-345e-472b-9d41-6e1ae58c4202:**@rabbit:5672/f561486f-803a-4612-ab79-d8fecb6badf1: failed to resolve broker hostname.
Trying again in 2.00 seconds...

[2022-03-28 14:38:17,438: ERROR/MainProcess] consumer: Cannot connect to amqp://26df2d3e-345e-472b-9d41-6e1ae58c4202:**@rabbit:5672/f561486f-803a-4612-ab79-d8fecb6badf1: failed to resolve broker hostname.
Trying again in 4.00 seconds...
```

Or, when I edit the broker URL by hand:

```
[2022-03-28 14:34:47,967: ERROR/MainProcess] consumer: Cannot connect to amqp://[...]:**@codabench.org:5672//: [Errno 99] Cannot assign requested address.
Trying again in 2.00 seconds...

[2022-03-28 14:34:58,097: ERROR/MainProcess] consumer: Cannot connect to amqp://[...]:**@codabench.org:5672//: [Errno 99] Cannot assign requested address.
Trying again in 4.00 seconds...
```

Didayolo:
Also, what is this for?

In .env:
```
# Location to store submissions/cache -- absolute path!
HOST_DIRECTORY=/your/path/to/codabench/storage
```

In the setup command:
```
-v /your/path/to/codabench/storage:/codabench \
```

Didayolo:
Edit: for a GPU worker, I tried to manually change rabbit to www.codabench.org

Didayolo:
@dtuantran @bbearce 

How about this? It is working now, right?

bbearce:
[image: image.png]

I see on FF that codabench.org is working. Is that what you meant? I also
see master is passing!!! :)

Didayolo:
@bbearce 

[...]

PR #725, opened 2022-10-12
Fix "rabbit" in BROKER URL, replace by domain name

Didayolo:
Fix for #690


Didayolo:
Replaced by #726 (cleaner)

PR #726, opened 2022-10-12
Use domain name in broker URL

Didayolo:
Fix for https://github.com/codalab/codabench/issues/690


Didayolo:
I've reverted it because it creates a bug. "An error occurred" when trying to create a queue.

Issue #738, opened 2022-11-09
No email address confirmation during sign up


Didayolo:
When creating an account, no confirmation email is sent, and it is not needed to start log in and using the account.

We need to implement that before starting to have more users.

bbearce:
Should be fixed. PR is: https://github.com/codalab/codabench/pull/751

PR #751, opened 2022-11-27
new feature: email confirmation

bbearce:
# @ mention of reviewers
@Tristan
@Didayolo 

There was no email confirmation and we don't want spam accounts to build up quickly. 

# Issues this PR resolves
https://github.com/codalab/codabench/issues/738

# Known issues to be addressed in a separate PR
...

- [x] Sign up and check to see that a green message shows up to let you know you need to check your email.
- [x] Check the confirmation link shows up and click it.
- [x] Verify that you are brought to the login page and see a message telling you that you are all signed up and to log in.
- [x] Check that you can log in.

Test error case:
- [x] Sign up and check to see that a green message shows up to let you know you need to check your email.
- [ ] Check the confirmation link shows up and copy most but not all of the link. 
- [ ] When navigating to this link, you should be brought to the sign-up page with red message indicating the link was corrupt. This should have deleted your user in the process and you should retry signing up. 
- [x] You should be able to correctly sign up, click activation link, and login after this.

...

# Misc. comments
I'm a little worried I unnecessarily added django "messages" (https://docs.djangoproject.com/en/4.1/ref/contrib/messages/) to the base.html template when I could up setup messages based on API calls (already how things are setup). The problem was I couldn't find how to catch the return response from the signup API call, which is where most messaging functionality happens. We didn't have messages setup as the code was using a javascript package called "toastr" to flash messages. I was able to use django messages (which is nice as we can use them now in the future) and wrap them in toastr so we can get the cool flashes that pop up and everything looks consistent. This is my only worry as far as implementing this correctly.

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [ ] Ready to merge


Didayolo:
I tried and it's working.

Some remarks:

- I needed to change "http" to "https" in the confirmation link to have it working. Maybe this is due to the test environment (Tuan is updating the SSL configuration) and will work fine in production

- If I try to login without having clicked on the link, I obtain this error message. It is the right behavior that it does not work, but the message is not explicit (should be saying "please confirm your email address").

Capture d’écran 2022-11-29 à 17 30 33

PR #777, opened 2023-03-14
Update .env_sample

Didayolo:


Didayolo:
@ihsaan-ullah @bbearce @acletournel 

This concerns the installation of Codabench. How can we confirm that this change is good?

If the installation previously worked for you, please try to re-install from scratch using this `.env`file.

Didayolo:
Not sure if this is useful

PR #816, opened 2023-04-21
Fix tests

Didayolo:
Wait around the click of the v15 submission E2E test. This should avoid some random failures of the CircleCI test. Also, the timeout was reduced.


Didayolo:
Maybe because the text is outside the blackbox for this competition?

Capture d’écran 2023-04-27 à 16 43 58

Issue #794, opened 2023-03-29
Flower can not start


dtuantran:
We get this error after deploying Codabench

![image](https://user-images.githubusercontent.com/7440668/228599212-fe1917f4-8cfb-4f24-96e2-9206b1b1a9e1.png)

As metioned here : https://stackoverflow.com/questions/64180054/importerror-cannot-import-name-command-from-celery-bin-base

It seems that our Flower version isn't compatible with Celery. Then we should check and fix those versions.

lbeziaud:
It works fine with python3.9 and celery<5.0.0

```diff
diff --git a/Dockerfile.flower b/Dockerfile.flower
index bce19a5..58d6540 100644
--- a/Dockerfile.flower
+++ b/Dockerfile.flower
@@ -1,10 +1,10 @@
-FROM python:alpine
+FROM python:3.9-alpine
 
 # Get latest root certificates
 RUN apk add --no-cache ca-certificates && update-ca-certificates
 
 # Install the required packages
-RUN pip install --no-cache-dir redis==3.0.1 flower==0.9.3
+RUN pip install --no-cache-dir redis==3.0.1 flower==0.9.3 "celery<5.0.0"
 
 # PYTHONUNBUFFERED: Force stdin, stdout and stderr to be totally unbuffered. (equivalent to `python -u`)
 # PYTHONHASHSEED: Enable hash randomization (equivalent to `python -R`)
```

PR #818, opened 2023-04-21
Flower with python3.9 and celery<5.0.0

Didayolo:
#794


Didayolo:
@dtuantran What do you think about this change?

Issue #722, opened 2022-10-11
Automatic tests randomly fail


Didayolo:
The CircleCI tests sometimes fail. They can be passed simply by re-running them, so it seems to be quite random.

Example (failure):

https://app.circleci.com/pipelines/github/codalab/codabench/887/workflows/8fcc2681-8b01-4f19-84eb-33a2d305da35/jobs/2554

After the re-run (success):

https://app.circleci.com/pipelines/github/codalab/codabench/887/workflows/7c377d7b-ce42-42d8-b8a0-82e02387d389/jobs/2556

The error is a time out exception on the following E2E selenium test:

```python
FAILED src/tests/functional/test_submissions.py::TestSubmissions::test_v15_submission_end_to_end
```

```python
/usr/local/lib/python3.8/site-packages/selenium/webdriver/support/wait.py:80: TimeoutException
```

bbearce:
I believe this is due to selenium requesting elements from the HTML DOM. It has to wait for them and sometimes it takes too long to get the elements. Sometimes not. We can continue to add sleep commands to fix this. 

More details...like javascript selenium requests objects and acts on them. Sometimes an element is a pop-up window. On that pop-up window could be a button selenium wants to click. If that pop-up hasn't loaded into selenium before it asks to click the button, selenium waits to get the button, but it never will as there was no button to click. Eventually waiting to click the button may time out.

Didayolo:
Interesting. Ideally we'd like to avoid these random failures because it is quite misleading. If adding some wait is sufficient to solve this then let's do it. I'm not sure if it is normal to need this or if it is bad practice.

bbearce:
It smells like bad practice and could need tweaking over time continuously. This may become an issue as codabench grows and or more tests are added.

The fundamental issue is that javascript can have its code written so that functions are forced to wait on each other to finish before carrying on. Maybe we can do the same with selenium and implement a "wait till the previous command finishes" versus a random tweaked sleep which may or may not be enough time.

bbearce:
I see this for instance:
https://www.selenium.dev/documentation/webdriver/waits/

Example:
```
driver.navigate("file:///race_condition.html")
WebDriverWait(driver, timeout=10).until(document_initialised)
el = driver.find_element(By.TAG_NAME, "p")
assert el.text == "Hello from JavaScript!"
```

The .until() is waiting for something. It still has a timeout but we can adjust that.

What we have is this:
[...]

PR #830, opened 2023-04-28
Remove the SNCF bundle e2e test

Didayolo:
The goal is to solve this issue:

- #722

With a previous PR (#828), **I added two v15 e2e tests: Iris code submission and Iris result submission**.

The test removed by this PR is testing in `test_submissions.py` "competition_15.zip" and "submission_15.zip", which is a template of SNCF competition. This bundle is not well rounded and documented, and the test seems to be randomly failing from time to time.

Maybe we want to really understand why this bundle creates problem. Or we can merge this PR and have stable tests from now on.

@dtuantran what do you think?


Didayolo:
I merge this but if we want we can keep a trace of the problem with the black box container.

Issue #874, opened 2023-05-12
Data inside workers


Didayolo:
In CodaLab, we have the possibility to add data directly inside the compute workers, so it can be read by the ingestion or scoring program:

https://github.com/codalab/codalab-competitions/wiki/User_Using-your-own-compute-workers#put-data-directly-inside-the-worker

The way it is implemented is simply that the folder `/data` of the worker is mounted on the docker container of the submission.

Do we have this on Codabench? If no, should we implement this in the same way than in CodaLab or is there a better practice?

dtuantran:
On Codabench we have `HOST_DIRECTORY=/codabench` env variable. I think we can do probably the same by putting data in `/codabench/storage/`. We need to try to see if it works.
Here is the code line related in compute_worker.py : https://github.com/codalab/codabench/blob/d8424b8c2a34fe485da52dd7e3dcee8b234bf6e5/compute_worker/compute_worker.py#L543

Didayolo:
Thanks for your input. If this work, I'll update the documentation here: https://github.com/codalab/codabench/wiki/Compute-Worker-Management---Setup

Didayolo:
To be confirmed:

- The data in `HOST_DIRECTORY` can be accessed
- The access is read-only

Issue #748, opened 2022-11-22
Detailed results not written to after 60 seconds, exiting!


Didayolo:
In CodaBench, if the `detailed_results.html` file is not created, the compute worker stop the execution (without logs on the platform side by the way).

I guess the idea behind this is to be able to have a look at the detailed results during the execution, in real time.

This is cool but not really robust. This causes error because usually this file is created at the end of the execution (especially in old CodaLab bundles).

See:

https://github.com/codalab/codabench/blob/232490ddf2682b89feedc2f6b907e88110828077/docker/compute_worker/compute_worker.py#L242

Update: if the `detailed_results.html` file is empty, it creates a bug.

**I think the file should be created by Codabench (the compute worker) itself if it does not exist.**

Didayolo:
Update: if the `detailed_results.html` file is empty, it creates a bug.

Didayolo:
The file should be created by CodaLab itself if it does not exist.

Didayolo:
I have this error again on another challenge:

```
Detailed results not written to after 60 seconds, exiting!
```

I think it is important that we fix it, because when the compute worker exit like this, the submission fails with no logs at all.

Didayolo:
Another take could be simply to remove this expiration of the scoring program.

LidiaGarrucho:
Hello,
Running the competition in a compute worker I still get:
Detailed results not written to after 60 seconds, exiting!
I have used the latest: codalab/competitions-v2-compute-worker:gpu1.1 in the compute worker.

LidiaGarrucho:
Using docker pull codalab/competitions-v2-compute-worker:gpu1.1 in the compute worker, I am getting:
Detailed results not written to after 60 seconds, exiting!

Why is this? Thank you!

LidiaGarrucho:
No, I am informing this error didn't dissapear when writing Detailed results running a competition on a compute worker. From this thread seems the scoring timeout was deleted but I am getting this message on the console when writing detailed results, even though at the Detailed results are submitted and displayed correctly in HTML format.

Didayolo:
> No, I am informing this error didn't dissapear when writing Detailed results running a competition on a compute worker. From this thread seems the scoring timeout was deleted but I am getting this message on the console when writing detailed results, even though the Detailed results are submitted and displayed correctly in HTML format.

It is actually a warning and not an error. I'll update the message so it is clearer (removing the "exiting" especially). Thank you for the report.

PR #881, opened 2023-05-17
Access data in worker

Didayolo:
Mount `$HOST_DIRECTORY/data` to `/app/data` in the container so it can be accessed by ingestion and scoring programs.

This can be used to store data directly inside the compute workers.

See:
- #874

Review:
- [x] Tested and it work
- [x] CircleCI tests are passing
- [x] Documentation is updated

Next steps:
- [x] Merge PR
- [ ] Update docker images (cpu and gpu)

### DIFF
--- a/compute_worker/compute_worker.py
+++ b/compute_worker/compute_worker.py
@@ -39,9 +39,9 @@ app.conf.task_queues = [
 
 
 # Setup base directories used by all submissions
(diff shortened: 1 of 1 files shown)


Didayolo:
@dtuantran we still need to update the docker images

Didayolo:
We should maybe fix this #748 before updating the dockers

Issue #713, opened 2022-09-29
Storage and clean up


Didayolo:
Implementing a storage quota and effectively managing capacity could be crucial for enhancing Codabench's scalability and ensuring long-term service availability in production.

# Features

### Deleting files from the MinIO

@OhMaley: Add a button, in admin analytics page, to manually starts a script removing all orphans files.

- [x] Nothing is never erased from the MinIO. Add a button to remove all orphan files manually (#1523, #1575)
- [x] To check: #1319

### Quota

We need to:
- [x] **Impose a storage quota to each user** (e.g. 15GB, platform setting) #1254
- [ ] Idea: separated quota for each joined competition 
- [x] Admins should be able to manually increase the quota of any user (on Django admin interface) #1254
- [x] Have an user interface to manage and delete submissions
Solved by: #918 
- Actually, submissions can't be removed if there are part of a benchmark (which makes sense to avoid spamming submissions when there is a daily limit).
- [x] Solution: **submissions are counted on the quota of the participants, but they can remove a submission (even if it's part of a benchmark) if it's not on the leaderboard.** #1738
- [x] Be able to delete tasks when deleting a benchmark #810
Partially Solved by: #918 
- [ ] _Be able to re-use past submissions_ #632
- [ ] Failed submission when deleted still remains as Data (deleted from leaderboard/submission panel but can be found in submissions in resources)
- [ ] The dumps should be counted too
- [ ] Refresh the quota figure automatically when submisison/dataset is uploaded in resource interface
- [ ] Readable format for the quota in the user table in admin interface .e.g MB

### Light/archive mode after competition end

Option to clean competitions when there are completed. It would for instance keep only leaderboard submissions.

### Automatic cleaning

- [ ] Failed submissions, orphans datasets, useless benchmarks, etc. should be marked and cleaned automatically. Users could manually unmark the object to keep it (from the resources interface). Deleted object should be kept in a trash can for one month. This feature needs to be discussed.
- [ ] #302 (schedule)

### Monitoring and statistics

- [x] Fix analytics page download CSV buttons: #649
Solved by: #988 
- [x] Improve analytics page to include storage statistics. Migrate analytics features from CodaLab.
Solved by: #1248 
- [x] date of storage calculation like codalab
Solved by: #1265 

### Per benchmark submission size limit

Migrate the CodaLab feature on Codabench. **This may be less important once the quota feature is implemented.**

- [ ] Limit the size of submissions
[...]

PR #895, opened 2023-05-26
File size added in readable format in Resources -> Datasets

ihsaan-ullah:
# @ mention of reviewers
@Didayolo 

Now users can see file size in readable format in Resources -> Datasets
Screenshot 2023-05-26 at 2 44 26 PM

# Issues this PR resolves
#713 -> Interfaces -> Point 1

> Add the size of the files

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/src/static/riot/datasets/management.tag
+++ b/src/static/riot/datasets/management.tag
@@ -80,7 +80,7 @@
                       Pagination
         -------------------------------------->
         <tr>
-            <th colspan="7" if="{datasets.length > 0}">
+            <th colspan="8" if="{datasets.length > 0}">
                 <div class="ui right floated pagination menu" if="{datasets.length > 0}">
                     <a show="{!!_.get(pagination, 'previous')}" class="icon item" onclick="{previous_page}">
                         <i class="left chevron icon"></i>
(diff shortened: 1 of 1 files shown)


Didayolo:
Thank you for the PR.

I think it breaks the buttons to navigate through pages in the resources interface (bottom right):

Capture d’écran 2023-05-26 à 15 38 43

Didayolo:
Nice work!

ihsaan-ullah:
@Didayolo 

When a submissions file is stored, the standard used for Kilobytes (KB) is 1024 and not 1000.
https://github.com/codalab/codabench/blob/develop/src/apps/datasets/models.py#L67

In the change I have made, I use 1000.
```python
// loop over n until it is greater than 1000
while(n >= 1000 && ++i){
     n = n/1000;
 }
```

Should this be changed to 1024?

Didayolo:
So I guess right now we showing MiB, GiB, etc. instead of MB or GB.

Yes, maybe it is better to change this to 1024.

ihsaan-ullah:
I think KiB, MiB,  GiB are multiples of 1024 while KB, MB, GB are multiples of 1000.

my change is showing KB, MB, Gb.

Didayolo:
OK then we can let it as it is, because the information provided is accurate.

Issue #909, opened 2023-06-01
Manual competition creation: fill in docker image by default


Didayolo:
When manually creating a benchmark, the field "Competition Docker Image" should contain `codalab/codalab-legacy:py37` by default:

<img width="964" alt="Capture d’écran 2023-06-01 à 17 08 55" src="https://github.com/codalab/codabench/assets/11784999/038b109a-8979-446b-8aab-d2fc0f5a1afb">

Issue #740, opened 2022-11-14
Saving a manually created competition with queue doesn't save


bbearce:
Below shows how if creating a competition and adding a queue, it doesn't save.

![image](https://user-images.githubusercontent.com/7649007/201777318-542cc26d-ea48-43d8-ad85-4fef96ee0fac.png)

Didayolo:
This is part of #733

PR #739, opened 2022-11-14
Issue 733 - Manual comp creation now will save leaderboard

bbearce:
# @ mention of reviewers
@Didayolo 
@Tthomas63 

# A brief description of the purpose of the changes contained in this PR.
Upon manual UI challenge creation, the leaderboard isn't saved.

# Issues this PR resolves
733 (https://github.com/codalab/codabench/issues/733)

# Known issues to be addressed in a separate PR
N/A

# A checklist for hand testing
- [ ] Make a challenge using the form in the UI and don't forget to create a leaderboard.
- [ ] Next once competition is created, go to edit and make sure the leaderboard you created still exists.
- [ ] Also go to results tab and make sure you can see the leaderboard.

# Any relevant files for testing
N/A

# Misc. comments
I'm worried that this fix and the underlying code I copied or updating a challenge is not what the original creators wanted. See this comment above the code I copied: https://github.com/codalab/codabench/blob/73cacee18a72ffa129938098b9b341d6b807759a/src/apps/api/views/competitions.py#L148.

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [ ] Code review by reviewer
- [ ] Hand tested by reviewer
- [ ] Ready to merge

PR #912, opened 2023-06-06
manual_comp_create_queue

bbearce:
This is code to fix the queue error when adding a competition manually with the editor. 

Manually creating a comp with queue throws wrong datatype error. We need to be able to set queue when creating competiton.

# Issues this PR resolves
- #740 
- #739 
- #909

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [ ] Ready to merge

### DIFF
--- a/src/apps/api/serializers/competitions.py
+++ b/src/apps/api/serializers/competitions.py
@@ -170,6 +170,9 @@ class CompetitionUpdateSerializer(CompetitionSerializer):
     phases = PhaseUpdateSerializer(many=True)
     queue = None
 
+class CompetitionCreateSerializer(CompetitionSerializer):
+    queue = None
+
 
 class CompetitionDetailSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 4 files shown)


Didayolo:
@bbearce Thanks for the PR.

There is no way to avoid the new class `CompetitionCreateSerializer` by having somehow the `queue = None` optionnaly in `CompetitionSerializer`?

I am testing the change right now.

bbearce:
Did you have a queue created? You need to have made one first. I imagine you did but not sure. Regarding the CompetitionCreateSerializer, there may be a way to just use the CompetitionSerializer and set queue = None after the fact. The CompetitionUpdateSerializer does the same pattern as CompetitionCreateSerializer. I copied the pattern.

Perhaps after instatiating the CompetitionSerializer we can do something like CompetitionSerializer.queue = None...

PR #934, opened 2023-06-12
Phase status fixed

### LINKED ISSUE #931: never ending phase has null status
Hi,

Phases without without end date have `"status": null` ([example](https://www.codabench.org/api/phases/1215/)). Apparently this *sometimes* breaks submission, with the submission form not being available to users. I have not experienced that (I'm an admin on tested competitions) but it was reported by two participants [here](https://www.codabench.org/forums/799/49/). This was "solved" on this particular challenge by setting an end date. Participant were not able to submit before, except on some unknown occasion (ending date was added on June 8, some submissions could be made on June 7, but reports were made after and before those submissions…).

The issue might be related to this line: https://github.com/codalab/codabench/blob/b43105a73a789a3223cf14b7c4c1640fddda1867/src/static/riot/competitions/detail/submission_upload.tag#L3 but a better fix would be to have `"status": "Current"` for phases without an end date. This could also be the result of an issue with the phase indexing.

Thank you @BrunetNathan for the initial report

ihsaan-ullah:
# @ mention of reviewers
@Didayolo 

Phases api (example: https://www.codabench.org/api/phases/1215/) was giving null status and status was not used in the UI to allow or block user from submiting to a phase which has started/ended etc.

Now this is fixed.

# Issues this PR resolves
#931 

# What is changed?

- [x] Phases api ( https://www.codabench.org/api/phases/) fixed to add a valid status instead of `null`
- [x] Used these 3 statuses  (1. `Previous`, 2. `Current`, 3. `Next`)
- [x] Serializer modified to return correct status
- [x] Phase start/end logic based on start date and end date replaced by status in the UI submission panel   
- [x] Admin as well as participants are allowed to see submission panel 
- [x] error/warning messages are displayed  to both admin and participant for not started and ended phases

# Screenshots
API with valid status:
Screenshot 2023-06-13 at 1 03 28 AM

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [ ] Code review by reviewer
- [ ] Hand tested by reviewer
- [ ] Ready to merge

### DIFF
--- a/src/apps/api/serializers/competitions.py
+++ b/src/apps/api/serializers/competitions.py
@@ -15,11 +15,13 @@ from profiles.models import User
 from tasks.models import Task
 
 from api.serializers.queues import QueueSerializer
+from datetime import datetime
 
 
 class PhaseSerializer(WritableNestedModelSerializer):
     tasks = serializers.SlugRelatedField(queryset=Task.objects.all(), required=True, allow_null=False, slug_field='key',
                                          many=True)
+    status = serializers.SerializerMethodField()
 
     class Meta:
         model = Phase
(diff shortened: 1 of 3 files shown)


ihsaan-ullah:
@Didayolo there is one problem with the status: It is never updated because there is no action for updating status so it always remains `null`

This cannot be done manually because there is no scenario where you can change status. Status depends on competition dates. Either this should be removed from the `Phase` Model or updated automatically by a cron job or something similar

PR #946, opened 2023-06-15
Use python3-setuptools

Didayolo:
Upgrade `python-setuptools` to `python3-setuptools` in the Dockerfile.


Didayolo:
I close this one because #947 is prefered.

Issue #1030, opened 2023-07-24
Highlight the primary column in the leaderboard


Didayolo:
Write the scores of the primary column of the task in bold, but only if there are multiple columns.

PR #1061, opened 2023-08-05
Leaderboard - bold score if column is primary

ihsaan-ullah:
# @ mention of reviewers
@Didayolo 

If a task has more than one scores, show the primary score (`index==0`) as bold text

Screenshot 2023-08-05 at 9 38 50 PM

# Issues this PR resolves
- #1030

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/src/apps/api/serializers/leaderboards.py
+++ b/src/apps/api/serializers/leaderboards.py
@@ -124,6 +124,7 @@ class LeaderboardPhaseSerializer(serializers.ModelSerializer):
     submissions = serializers.SerializerMethodField(read_only=True)
     columns = serializers.SerializerMethodField()
     tasks = PhaseTaskInstanceSerializer(source='task_instances', many=True)
+    primary_index = serializers.SerializerMethodField()
 
     def get_columns(self, instance):
         columns = Column.objects.filter(leaderboard=instance.leaderboard, hidden=False)
@@ -131,6 +132,9 @@ class LeaderboardPhaseSerializer(serializers.ModelSerializer):
             raise serializers.ValidationError("No columns exist on the leaderboard")
         else:
             return ColumnSerializer(columns, many=len(columns) >= 1).data
+
+    def get_primary_index(self, instance):
+        return instance.leaderboard.primary_index
 
     class Meta:
         model = Phase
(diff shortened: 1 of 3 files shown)


Didayolo:
@ihsaan-ullah Is the primary column always at index 0?

Given that you can select any column as primary using the editor:

Capture d’écran 2023-08-09 à 12 52 54

ihsaan-ullah:
> @ihsaan-ullah Is the primary column always at index 0?
> 

Yes, primary column means index = 0
This can be seen here (although never explained)
https://github.com/codalab/codabench/blob/82b03a497d131487b770e6538a4e863f20455ccb/src/static/riot/competitions/editor/_leaderboard.tag#L404

Didayolo:
@ihsaan-ullah Maybe it is not a problem, or at least could be for a separate PR:

When the other columns are hidden, it still acts as if there are multiple columns and put the scores in bold.

ihsaan-ullah:
> @ihsaan-ullah Maybe it is not a problem, or at least could be for a separate PR:
> 
> When the other columns are hidden, it still acts as if there are multiple columns and put the scores in bold.

Can you share a screenshot of this problem. I think this should not be happening because at some point the we have changed the server to not return hidden columns. But this can be investigated further

Didayolo:
Actually it seems that even when I remove the other column, the score is still in bold. I think the problem is actually that the number of scores is fetched from the submission, not from the leaderboard, so you need to upload a new submission to update this.

Another problem with that is that if you edit the leaderboard to change the primary column, the previous submission will still be in bold on the previous primary column.

ihsaan-ullah:
@Didayolo 

With latest push, the following will happen

- Bold score is now linked to leaderboard. if you change the primary column from editor, bold columns will change
- Bold column will appear only if you have more than one columns showing in the leaderboard
- There was a bug that the hidden column scores were returned from backend. now if you hide a column, its score will not be returned to front end

Didayolo:
Thank you for your quick update. I'll test it again asap.

Didayolo:
@ihsaan-ullah 

Is it normal that you have added `index` here? Now it is twice:

Capture d’écran 2023-08-10 à 16 49 51

ihsaan-ullah:
> @ihsaan-ullah
> 
> Is it normal that you have added `index` here? Now it is twice:
> 
> Capture d’écran 2023-08-10 à 16 49 51

this was by mistake. just removed it in the latest push

Issue #744, opened 2022-11-16
We need more information in job status page


Didayolo:
Here is the interface:

<img width="1153" alt="Capture d’écran 2022-11-16 à 14 17 05" src="https://user-images.githubusercontent.com/11784999/202191385-defb0b91-cc08-422e-9298-8687a7f31a38.png">

And the old one in CodaLab Competitions:

<img width="1158" alt="Capture d’écran 2022-11-16 à 14 18 20" src="https://user-images.githubusercontent.com/11784999/202191395-f897ed7a-066a-4894-8b7e-95d307aec7d9.png">

The queue (custom queue name or * for default) and the hostname (container ID) are especially important.

TODO:

- [x] **Hostname of both ingestion and scoring jobs**
Solved by: #1126 
- [x] Hostname
Solved by: #861
- [x] Make competition names as clickable links
Solved by: #897
- [x] **Queue name**
Solved by: #1091 
- [x] Have a way to differentiate and show/hide the "secondary submissions", which are hidden submissions made when there are multiple tasks
Solved by: #1096 
- [x] Add queue in submission
Solved by: #1114 

New feature - All users should be able to access the job status interface:
- [x] Admins should be able to see all queues (including the default queue)
- [x] users should be able to see status page with their own submissions only
Solved by: #1098 
- [x] Users should be able to see only the queue they are administrating. This is indeed useful when debugging and monitoring a competition
Solved by: #1098 
- [x] When this is done, we should add a button in the Queue Management page that redirects to the job status page.
Solved by: #1098 

Other TODOs (less important):
- [x] Make usernames as clickable links
Solved by #1092
- [x] Submission size
Solved by #1093
- _Run time (see also #610) (abandoned)_

ihsaan-ullah:
@Didayolo what do you want to get here:

> Have a way to differentiate and show/hide the "secondary submissions", which are hidden submissions made when there are multiple tasks

From this screenshot
<img width="1130" alt="Screenshot 2023-08-20 at 1 32 54 PM" src="https://github.com/codalab/codabench/assets/13259262/5ef4ff20-3285-4dcf-be6f-f5088e61f290">
You can see that submission id 2 is the actual submission but id 3,4 are generated because this competition has 2 tasks. 
Do you want to show only submission id 2 and not 3,4?

ihsaan-ullah:
@Didayolo 

Can you please explain these two points:

> Users should be able to see only the queue they are administrating
[...]

PR #1098, opened 2023-08-21
Server status - Show to all users

ihsaan-ullah:
# @ mention of reviewers
@Didayolo 

Now all users can see `server_status` page. 
Everything stays the same for super admin but for normal users they can do the following:

- see their own submissions
- see the submissions which are using a queue whose owner is this user.

Note: a user can see past submissions too if you change a queue in a competition from default to this user's queue. And user can lose all submissions showing before but not anymore because queue is changed to another queue. This happens because `queue` is part of a competition and not a submission.

# Screenshots:
status page link in the menu:
Screenshot 2023-08-21 at 6 25 06 PM

user `ihsan01` can see his own submissions running on default queue (*) and he can also see some other submissions from other user because they are running on queue: `aaa` whose owner is `ihsan01`
Screenshot 2023-08-21 at 6 23 36 PM

Button in queue page to go to
Screenshot 2023-08-21 at 6 43 09 PM
 server status page

# Issues this PR resolves
- #744 -> users should be able to see status page with their own submissions only
- #744 -> Users should be able to see only the queue they are administrating. This is indeed useful when debugging and monitoring a competition
- #744 -> When this is done, we should add a button in the Queue Management page that redirects to the job status page.

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/src/apps/pages/views.py
+++ b/src/apps/pages/views.py
@@ -59,15 +59,29 @@ class ServerStatusView(TemplateView):
     template_name = 'pages/server_status.html'
 
     def get_context_data(self, *args, **kwargs):
-        if not self.request.user.is_staff:
-            raise HttpResponse(status=404)
 
         show_child_submissions = self.request.GET.get('show_child_submissions', False)
 
+        # Get all submissions
         qs = Submission.objects.all()
+
+        # If user is not super user then:
+        # filter this user's own submissions
+        # and
+        # submissions running on queue which belongs to this user
+        if not self.request.user.is_superuser:
+            qs = qs.filter(
+                Q(owner=self.request.user) |
+                Q(phase__competition__queue__isnull=False, phase__competition__queue__owner=self.request.user)
+            )
+
+        # filter for fetching last 2 days submissions
         qs = qs.filter(created_when__gte=now() - timedelta(days=2))
+
+        # filter out child submissions i.e. submission has no parent
         if not show_child_submissions:
             qs = qs.filter(parent__isnull=True)
+
         qs = qs.order_by('-created_when')
         qs = qs.select_related('phase__competition', 'owner')
 
--- a/src/apps/pages/views.py
+++ b/src/apps/pages/views.py
@@ -1,6 +1,4 @@
 from datetime import timedelta
-
-from django.http import HttpResponse
 from django.utils.timezone import now
 from django.views.generic import TemplateView
 from django.db.models import Count, Q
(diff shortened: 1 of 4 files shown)


Didayolo:
It is not due to this PR, because I can reproduce the bug in `develop`, but I have identified a problem.

Steps to reproduce:
- Create a queue "Test 1"
- Link the queue "Test 1" to a competition
- Make a submission (no worker is linked to the queue, so it will stay in "submitting" state)
- Change the queue of the competition to another one, "Test 2"
- Make a new submission
- Now, in the server status, the two submissions are marked as submitted to the queue "Test 2"

ihsaan-ullah:
> @ihsaan-ullah
> 
> It is not due to this PR, because I can reproduce the bug in `develop`, but I have identified a problem.
> 
> Steps to reproduce:
> 
> * Create a queue "Test 1", and link it to a competition
> * Make a submission
> * Change the queue of the competition to another one, "Test 2"
> * Now, in the server status, the two submissions are marked as submitted to the queue "Test 2" (and the owner of the queue "Test 2" can see both)
> 

I agree with this. This is what I mentioned in the Note in the PR:

**Note: a user can see past submissions too if you change a queue in a competition from default to this user's queue. And user can lose all submissions showing before but not anymore because queue is changed to another queue. This happens because queue is part of a competition and not a submission.**

> We just have to change the queue, without re-submitting, and the queue is changed for all submissions:

We are not changing the queue of submission. The problem is that queue is not linked to a submission but to a competition. We can get the queue from competition only.

> The problem may come from this PR #1091

No this PR is just displaying thename of the queue of the competition which has this submission.

> The way the queue is taken basically just reads the queue of the competition, and not the queue in which the submission was done?

Yes you are right. I think there is no way to track the queue of submission. I maybe wrong but I did not find it anywhere

Didayolo:
@ihsaan-ullah That's interesting. I am not really concerned about the potential leakage of information when you switch the queue; if the competition is using your queue, then you are most probably one of the competition organizers. Also, this interface does leak very sensitive information such as IP address of the workers.

The main problem is more that the table is a bit confusing. When trying to debug submissions, if you are not aware of this behavior, it can be misleading.

ihsaan-ullah:
> @ihsaan-ullah That's interesting. I am not really concerned about the potential leakage of information when you switch the queue; if the competition is using your queue, then you are most probably one of the competition organizers. Also, this interface does leak very sensitive information such as IP address of the workers.
> 
> The main problem is more that the table is a bit confusing. When trying to debug submissions, if you are not aware of this behavior, it can be misleading.

We should discuss this in the meeting to plan something for it

Issue #968, opened 2023-06-20
Uploading a scoring program with the directory structure


Didayolo:
When you upload a scoring program or an ingestion program in your `Resources`, you can either zip it with the directory structure:

```
scoring_program.zip
|-- scoring_program/
|---|-- score.py
|---|-- metadata
```

or without the directory structure:

```
scoring_program.zip
|-- score.py
|-- metadata
```

I think that right now only the second option works. We should make this more robust.

PR #1121, opened 2023-08-25
Compute Worker - Scoring program with dir structure

ihsaan-ullah:
# @ mention of reviewers
@Didayolo 

# Note:
This a critical PR as it changes the Compute worker. I have added comments to the added code but if that is not sufficient, more explanation should be added. The only part of the code touched is when the `program` is extracted while scoring. I checked and when `destination == "program"` and `is_scoring` is `True` then this means that scoring program is being extracted. We may need the same functionality i.e. extract with or without parent directory for other programs too (to be tested and discussed)

Also note that we can identify a scoring program running by `self.is_scoring` in the compute worker

Now a scoring program with directory structure will work in the compute worker to execute a submission
```
scoring_program.zip
|-- scoring_program/
|---|-- score.py
|---|-- metadata
```

# Issues this PR resolves
- #968 

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [ ] Code review by reviewer
- [ ] Hand tested by reviewer
- [ ] Ready to merge

### DIFF
--- a/compute_worker/compute_worker.py
+++ b/compute_worker/compute_worker.py
@@ -362,7 +362,47 @@ class Run:
 
         # Extract the contents to destination directory
         with ZipFile(bundle_file, 'r') as z:
-            z.extractall(os.path.join(self.root_dir, destination))
+
+            # Run for scoring program only
+            if destination == "program" and self.is_scoring:
+
+                # Check if scoring program has a parent directory
+                # or the files are zipped without parent directory
+                # If parent directory is found i.e. parent_dir is not None
+                # then extract the scoring program and move its content to
+                # destination dir and then delete the parent dir
+
+                # Get list of files and directories from the zip of scoring program
+                extracted_files = z.namelist()
+
+                # Intiialize parent dir to none
+                parent_dir = None
+
+                # loop over all the extracted files
+                # to identify parent directory
+                # the first directory found is considered parent dir
+                for extracted_file in extracted_files:
+                    # check if a metadata file is located in the subdirectory
+                    # that directory is considered the parent dir
+                    if '/' in extracted_file and os.path.basename(extracted_file) == 'metadata':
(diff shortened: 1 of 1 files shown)


compute_worker/compute_worker.py
```diff
@@ -362,7 +362,51 @@ def _get_bundle(self, url, destination, cache=True):
...
+                # Intiialize parent dir with a None value
+                parent_dir = None
+
+                # loop over all the extracted files to identify parent directory
+                for extracted_file in extracted_files:
+                    # if a metadata file is located in the subdirectory
+                    # that directory is considered the parent dir
+                    # Note:
+                    # `/` shows that there is a directory structure e.g. scoring_program/metadata
+                    if '/' in extracted_file and os.path.basename(extracted_file) == 'metadata':
```
ihsaan-ullah, line 388: @Didayolo here i have used condition of `metadata` but I think we can have different metadata files like metadata.yaml

We should discuss this

Didayolo: Yes, you can see here that metadata file can be `metadata` or `metadata.yaml`:

https://github.com/codalab/codabench/blob/d2f1197b79b1e613214297f0f2404cea11c9251d/compute_worker/compute_worker.py#L527C9-L527C9

ihsaan-ullah: > Yes, you can see here that metadata file can be `metadata` or `metadata.yaml`:
> 
> https://github.com/codalab/codabench/blob/d2f1197b79b1e613214297f0f2404cea11c9251d/compute_worker/compute_worker.py#L527C9-L527C9

Adding a condition of metadata.yaml

Didayolo:
- I don't think this change should be specific to the scoring program
- We made a lot of changes to `compute_worker.py` since the creation of this PR

PR #1140, opened 2023-09-04
Compute worker - show docker pull image failure logs

ihsaan-ullah:
I have added some code to show docker pull failure logs in the ingestion logs. But there are some problems 

1. When you make the function `_get_container_image` async, then I am able to write to the logs BUT then the code does not stop and moves to the scoring program which is expected because we don't use `await` to wait for _get_container_image to finish its execution.
2. If you remove async from `_get_container_image` then you get an error that the logs files does not exist 

This needs some discussion 

***

this is how we see the logs:
Screenshot 2023-09-05 at 12 04 02 AM
Screenshot 2023-09-05 at 12 04 46 AM

### DIFF
--- a/compute_worker/compute_worker.py
+++ b/compute_worker/compute_worker.py
@@ -360,7 +360,7 @@ class Run:
         #     })
         self._update_submission(data)
 
-    def _get_container_image(self, image_name):
+    async def _get_container_image(self, image_name):
         logger.info("Running pull for image: {}".format(image_name))
         try:
             cmd = [CONTAINER_ENGINE_EXECUTABLE, 'pull', image_name]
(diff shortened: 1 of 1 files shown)


Didayolo:
That is nice, also the cleaning of the code. 

I'm wondering if adding `logger` helps in having logs in the platform's interface.

We can discuss this at the meeting.

ihsaan-ullah:
> That is nice, also the cleaning of the code.
> 
> I'm wondering if adding `logger` helps in having logs in the platform's interface.
> 
> We can discuss this at the meeting.

No, to see logs we need to add them to std_out or std_err of ingestion or scoring.

Didayolo:
Suggestion: having a new log file specifically for compute worker logs

ihsaan-ullah:
A new PR is on the way for this issue

PR #1144, opened 2023-09-08
Server status - Rabbit MQ and Flower links fixed

ihsaan-ullah:
# @ mention of reviewers
@Didayolo 

# A brief description of the purpose of the changes contained in this PR.
Now Rabbit MQ and Flower links point to the right URL as mentioned here: https://github.com/codalab/codabench/wiki/Administrator-procedures

# Issues this PR resolves
- #1105 

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] CircleCi tests are passing
- [x] Ready to merge

PR #1145, opened 2023-09-08
Server status - flower and rabbit mq links

ihsaan-ullah:
Used `DOMAIN_NAME` instead of `SITE_DOMAIN`

Linked to : #1144


Didayolo:
Now the URLs are `codabench-test.lri.fr:15672` and `codabench-test.lri.fr:5555` but it does not resolve

PR #1148, opened 2023-09-09
Server status - Rabbit and flower links

ihsaan-ullah:
`https://` added with domain name. Now it should work.
Locally `https` does not work but `http` does. I am keeping https because this feature is not useful locally

Linked to : 
- #1144 
- #1145


Didayolo:
Still not loading for some reason... We need first to find links to RabbitMQ and Flower in the test server that does work

PR #1168, opened 2023-09-20
Update docker-compose.yml

dtuantran:
Done on production


Didayolo:
So it is done on both production and test server, right?

When changing this on local instance, is there any actions to make?

Issue #1171, opened 2023-09-22
Leaderboard not well ordered


Didayolo:
The leaderboard the validation phase of the following competition is not well ordered:

https://www.codabench.org/competitions/1219/

![20230922](https://github.com/codalab/codabench/assets/11784999/44b155ca-68cc-44f2-aa2c-e53aac055f99)

(the competition is private, if you need the secret URL please ask me)

---

UPDATE:

This problem is linked to another problem: the score in the tab "Submissions" (either participant tab or admin tab) does not reflect perfectly the score in the leaderboard.

This is likely due to the javascript rounding found in [submission_manager.tag](https://github.com/codalab/codabench/blob/develop/src/static/riot/competitions/detail/submission_manager.tag):

```javascript
        self.get_score = function (submission) {
            try{
                return parseFloat(submission.scores[0].score).toFixed(2)
           [...]
```

It seems that javascript standard for floats is not precise. The order is based on the scores that you can find in the "Submissions" tab, not the actual scores. The leaderboard reflect the true score.

---

So, it is ordered by primary column, but using a bad rounding.

Didayolo:
Even clicking on "Score" to re-order the rows does not give the right order

ihsaan-ullah:
I think this is not competition specific problem.

In this competition: https://www.codabench.org/competitions/984/#/results-tab
you can see there is a problem 

DESC
<img width="215" alt="Screenshot 2023-09-22 at 5 25 34 PM" src="https://github.com/codalab/codabench/assets/13259262/d1352343-e38b-4a1b-85b4-e8043fcc3bd6">

ASC
<img width="213" alt="Screenshot 2023-09-22 at 5 25 25 PM" src="https://github.com/codalab/codabench/assets/13259262/756f9072-74ea-4bb9-8442-f333f7ba1614">

Didayolo:
I don't understand where the order comes from. It could be based on another score, or on the best score of the participant, maybe on the date... Maybe we should investigate the code directly.

ihsaan-ullah:
> I don't understand where the order comes from. It could be based on another score, or on the best score of the participant, maybe on the date... Maybe we should investigate the code directly.

When you load the leaderboard, the order seems correct. I think the order is changed and is buggy because of the reordering when you click the column 
[...]

PR #1186, opened 2023-10-05
Order leaderboard only by primary column and date of creation

Didayolo:
It seems that the ordering does not follow the primary column sometimes. This code change should ensure that only the primary column, and secondly date of creation, are used to order submissions on the leaderboard.

We need a competition bundle and submissions on which we can easily reproduce the wrong order bug to test this.

# Issues this PR resolves

- #1171

# Checklist
- [x] Code review by me 
- [ ] Hand tested by me 
- [x] I'm proud of my work
- [ ] Code review by reviewer
- [ ] Hand tested by reviewer
- [ ] Ready to merge

### DIFF
--- a/src/apps/api/serializers/leaderboards.py
+++ b/src/apps/api/serializers/leaderboards.py
@@ -100,22 +100,20 @@ class LeaderboardEntriesSerializer(serializers.ModelSerializer):
         # desc == -colname
         # asc == colname
         primary_col = instance.columns.get(index=instance.primary_index)
-        # Order first by primary column. Then order by other columns after for tie breakers.
-        ordering = [f'{"-" if primary_col.sorting == "desc" else ""}primary_col']
+        # Order first by primary column. Then order by date of creation for tie breakers.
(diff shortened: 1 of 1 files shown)


bbearce:
I made submissions that randomly create numbers between 0 and 1. That is the main column. I also changed my computer's date halfway through making submissions for participants. Then I made sure 2 different submissions on different dates by different participants had the exact same score. Here are the results:

image

I believe it is behaving correctly and sorts desc for the correct col and users bbearce and guest5 have the exact same score on different days. It gives preference to the earlier time.

bbearce:
However I think there is a problem where the second column is being used for ordering. See this example:

image

ihsaan-ullah:
Leaderboard score has sometimes problems. Here is a screenshot from a recent competition where lower score can be seen ranked higher

I am not sure if this PR solves this issue too

image_2024_03_08T13_16_13_503Z

Didayolo:
Thanks for your input. I am not sure too if this PR helps.

Issue #1281, opened 2024-01-17
[New Feature] Toggle Auto-run submissions


ihsaan-ullah:
Provide a toggle option to organizers in competition settings, to enable/disable auto-run submissions.

- By default auto-run submissions will be enabled
- When this is disabled, all submissions submitted will have run submission button infront of them and by clicking this button, organizers can run the submission.

Didayolo:
I do not understand. When the auto-rerun going to be triggered?

ihsaan-ullah:
by default now we have auto-run which means that when you submit a submission it is added to the queue. When you turn this off (in this feature) then organizer can manually run it

PR #1294, opened 2024-01-23
Competition Auto-Run Submissions

ihsaan-ullah:
# @ mention of reviewers
@Didayolo 

Organizers can enable or disable `auto_run_submissions` feature from competition setting to allow participant's submissions to auto-run (the default way) or to require an organizer to run the submission manually (the new feature)

### How to set `auto_run_submissions` 
1. You can set it in the competition settings:
Screenshot 2024-01-23 at 11 07 12 AM

2. You can set it in the yaml file:
Screenshot 2024-01-23 at 11 07 37 AM

Note: default value is `True` so that the 

### How do participants know their submission are not auto-run?
Participants see a question mark with a message to tell them that organizers will run their submissions:
Screenshot 2024-01-23 at 11 10 38 AM

### How can organizer run the submission?
Organizer can see a `play` button for submissions which should be executed manually. By clicking this button, the submission will run as a normal submission
Screenshot 2024-01-23 at 11 11 19 AM

# Issues this PR resolves
- #1281 

- [x] Create a competition and disable `auto_run_submissions` either from competition setting or from competition yaml
- [x] Submit a submission and check if you see a message that the submission will be run by organizer
- [x] As an organizer, check that you see the play button in front of the submission and submission runs when you click it
- [x] upload two competitions one with `auto_run_submission` set to true and other with false In the yaml file to see they are correctly shown in the competition setting
- [x] Enable/Disable `auto_run_submissions` from the setting to check that uploaded submissions follow the setting

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [ ] Ready to merge


cjh1:
Quick question, what is the difference between "Submitting and Submitted". To me I would expect as submission to be in the  "Submitted" state after I have uploaded my submission?

src/apps/api/serializers/submissions.py
```diff
@@ -41,6 +41,7 @@ class SubmissionSerializer(serializers.ModelSerializer):
     on_leaderboard = serializers.BooleanField(read_only=True)
     task = TaskSerializer()
     created_when = serializers.DateTimeField(format="%Y-%m-%d %H:%M")
+    auto_run = serializers.SerializerMethodField(read_only=True)
```
cjh1, line 44: Why do we need to add `auto_run` to every submission? To me if you want to see if a submission should be auto run, you should check the associated competition?

src/apps/competitions/models.py
```diff
@@ -538,6 +542,13 @@ def start(self, tasks=None):
         from .tasks import run_submission
         run_submission(self.pk, tasks=tasks)
 
+    def run(self):
+        # get tasks from the phase
+        tasks = self.phase.tasks.all()
+        # start submission providing the tasks
+        self.start(tasks=tasks)
+        return self
+
```
cjh1, line 551: Why is this change needed?

ihsaan-ullah:
> Quick question, what is the difference between "Submitting and Submitted". To me I would expect as submission to be in the "Submitted" state after I have uploaded my submission?

The first status is`Submitting` when you upload a submission. Submitted is used when submission is forwarded for running

src/apps/api/serializers/submissions.py
ihsaan-ullah: This is added so that it can be accessed easily because submission is not directly linked to a competition but through a phase

src/apps/competitions/models.py
ihsaan-ullah: this change is for running the submission inside run_submission function. self.start forwards the submission for running

Issue #930, opened 2023-06-11
email all participants markdown to html broken


lbeziaud:
Hi,

The formatting of emails sent through the "email all participants" functionality is broken. HTML tags are escaped. For example, sending:
```markdown
**bold** [link](https://codabench.org)
```
results in:
```
--===============0869688155652953062==
Content-Type: text/plain; charset="utf-8"
MIME-Version: 1.0
Content-Transfer-Encoding: 7bit

<p>&lt;strong&gt;bold&lt;/strong&gt; &lt;a href="https://codabench.org"&gt;link&lt;/a&gt;</p>

--===============0869688155652953062==
Content-Type: text/html; charset="utf-8"
MIME-Version: 1.0
Content-Transfer-Encoding: 7bit

<p>&lt;strong&gt;bold&lt;/strong&gt; &lt;a href="https://codabench.org"&gt;link&lt;/a&gt;</p>
--===============0869688155652953062==--
```
which renders as:
```
<strong>bold</strong> <a href="https://codabench.org">link</a>
```
rather than
> **bold** [link](https://codabench.org)

ihsaan-ullah:
<img width="635" alt="Screenshot 2023-06-12 at 1 53 21 AM" src="https://github.com/codalab/codabench/assets/13259262/262dec65-6536-4b55-9694-13b0f17a771e">

ihsaan-ullah:
> Thank you for reporting this issue. Is it related to this change or is it a separate part of the code?
> 
> * [html is now interpreted in compettion pages #878](https://github.com/codalab/codabench/pull/878)

This seems to be a separate issue. It looks like it is never tested with HTML
This function receives raw html and while sending,  the html is not properly parsed (needs investigation)
https://github.com/codalab/codabench/blob/e7de84f703a5a56636800f0545be538795560194/src/utils/email.py#LL42C5-L42C32

Didayolo:
> How can this be tested locally?

Maybe you can try "to send an email to participants" locally and then find the raw body of the email inside the Django logs using `docker-compose logs django`.

lbeziaud:
You might not need that. The form sends the data as HTML to the API, rather than in plain markdown as requested by https://github.com/codalab/codabench/blob/b728d062712e09e462f8cf422ce90a8166c79244/src/apps/competitions/tasks.py#L743C33

ihsaan-ullah:
Solved by #1340

PR #1340, opened 2024-02-17
html tags in email text

ihsaan-ullah:
@Didayolo 

we need to test this on test server. I think this change should fix the issue

### How to test:

Send an email to a participant with text:
```
**bold** [link](https://codabench.org)
```

Participant should receive:

**bold** [link](https://codabench.org)

Participant should not receive:
```
<strong>bold</strong> <a href="https://codabench.org">link</a>
```

### Why I think this is fixed?

Before the fix, I see the following in the django logs:

```
<p>&lt;strong&gt;bold&lt;/strong&gt; &lt;a href="https://codabench.org"&gt;link&lt;/a&gt;</p>
```

after the fix I see this:
```
<p><strong>bold</strong> <a href="https://codabench.org">link</a></p>
```

If this fixes the problem, then I will cleanup and add some comments

### DIFF
--- a/src/utils/email.py
+++ b/src/utils/email.py
@@ -41,9 +41,10 @@ def sanitize(content):
 
 def codalab_send_markdown_email(subject, markdown_content, recipient_list, from_email=None):
     from_email = from_email if from_email else settings.DEFAULT_FROM_EMAIL
-    message = sanitize(markdown_content)
-    html_message = sanitize(markdown.markdown(message))
+    html_message = markdown.markdown(markdown_content)
+    # message = sanitize(markdown_content)
+    # html_message = sanitize(markdown.markdown(message))
 
-    message = EmailMultiAlternatives(subject, message, from_email=from_email, bcc=recipient_list)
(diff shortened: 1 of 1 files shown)


Didayolo:
The mail service from the test server seems to have a problem.

- I sent an email from codabench.org it worked
- I sent an email from codabench-test.lri.fr it did not work

Didayolo:
Linked issue: 

- #930

Didayolo:
@ihsaan-ullah 

With `noreply@codabench-test.lri.fr` it does not work:

Capture d’écran 2024-02-22 à 17 00 11

ihsaan-ullah:
I think we need to create this email `noreply@...` associated with the domain `codabench-test...` 

But we can still test this fix right?

Didayolo:
Yes, we can still test the fix. I'll do it. The mailing problem is simply a configuration problem of the test server.

Didayolo:
Tested on production and it works!

ihsaan-ullah:
> Tested on production and it works!

nice. I am closing the related issue

Issue #1180, opened 2023-10-03
Manual Competition Creation has bug with pages not able to be shown


bbearce:
# Overview
Manual Competition Creation finishes the creation and saves the competition, but the pages have a bug where you can't select them.

## Own Instance
In my instance I created a competition with pages and info. However only the Terms and Files tabs work after I select anything other than Overview. After this even Overview won't work:

![custom_challenge_tabs_not_working](https://github.com/codalab/codabench/assets/7649007/c796751b-f38d-42ee-b631-d20a9948922a)

## Did the same on Prod
[Comp Ex](https://www.codabench.org/competitions/1445/)
I have the same issue where Overview won't work after clicking off of it.
![comp_creation_tabs_master_branch](https://github.com/codalab/codabench/assets/7649007/dccdb588-fba4-49c0-81a3-7ce976cefe33)

## Challenge Creation Details:
![comp_creation_details](https://github.com/codalab/codabench/assets/7649007/bb10d18e-a792-4f5f-964d-83624151adca)
![comp_creation_pages](https://github.com/codalab/codabench/assets/7649007/0d63597f-8d0e-4a02-931d-a059aafc5835)
![comp_creation_phases](https://github.com/codalab/codabench/assets/7649007/710a4add-6962-40bb-bc88-4693fe6ed9b8)
![comp_creation_leaderboard](https://github.com/codalab/codabench/assets/7649007/577e5e69-e1c4-4f34-9890-f42a06332900)
![comp_creation_administrators](https://github.com/codalab/codabench/assets/7649007/0d906d20-4277-44c2-b560-d599158f77da)

bbearce:
Yeah, We can always reopen if I'm wrong. Better to be lean.

ff98li:
Hi, I am stuck with exactly the same issue. I have tried opening the competition page in different browsers on different machines and none of the page tabs are clickable (except for terms). May I ask how you resolved this?

ff98li:
> So upon further investigation I cannot reproduce: https://www.codabench.org/competitions/1676/ https://www.codabench.org/competitions/1445/

@Didayolo @bbearce Upon further investigation, I have figured out the cause. This bug would be triggered when the current date is not within a competition's phase. So the page tabs in the above instances are no longer clickable at this point. You might consider reopening this issue?

Didayolo:
Apperently is not only a problem with manual competition creation. Ihsan will upload an example bundle.

bbearce:
[...]

PR #1344, opened 2024-02-21
Manual UI Creation Bug Fix (soome pages not rendering)

bbearce:
# Issue description
Some pages aren't showing if you are not in a phase.

# Issues this PR resolves
- #1180

- [x] Upload competition
- [x] Delete all but one phase
- [x] Make sure you can access all html pages
- [x] Repeat for pre-existing and manually created challenges

[iris_competition_bundle_vetted.zip](https://github.com/codalab/codabench/files/14358178/iris_competition_bundle_vetted.zip)

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/src/apps/api/serializers/competitions.py
+++ b/src/apps/api/serializers/competitions.py
@@ -24,6 +24,7 @@ class PhaseSerializer(WritableNestedModelSerializer):
     tasks = serializers.SlugRelatedField(queryset=Task.objects.all(), required=True, allow_null=False, slug_field='key',
                                          many=True)
     status = serializers.SerializerMethodField()
+    is_final_phase = serializers.SerializerMethodField()
 
     class Meta:
         model = Phase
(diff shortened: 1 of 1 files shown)


ihsaan-ullah:
What was the actual problem?

bbearce:
So elusive man. Phases have these two types of statuses. status and is_final_phase. Generally speaking if the phase status is current then everything works out fine. In our case is there is no current phase and there is some JavaScript trying to look up in that case if the is_final_phase of any phase in this challenge is the final phase. That part wasn't working because even if you only have one phase there was no guarantee it was the final. To me it was a back end bug where if a challenge only has one phase it needs to always say that it is the final phase in the data base. It wasn't doing that and it was getting kind of tricky for me to figure out. So instead of hacking the front end to just figure out if there's only one phase and force the issue, I decided to change the back end to adjust the phase's attribute appropriately.

Didayolo:
@bbearce By default `is_final_phase` is False? Where is it defined?

ihsaan-ullah:
I think this 'is_final_phase' is not needed. We can remove it from the model and remove this logic. But maybe in another PR. The db design is not what i would do in this case

Issue #1451, opened 2024-05-22
Storage Analytics scan failing + wrong dates


ObadaS:
Hello,
I have 2 problems related to Storage Analytics to report.

### **Problem 1:**

- [x] Solved #1465

There seems to be a problem with the Storage Analytics function.

When the function is run, it fails immediately with the following error :

```python
[2024-05-15 08:50:42,369: INFO/MainProcess] Received task: analytics.tasks.create_storage_analytics_snapshot[fb0db7bd-5e81-4e1c-9946-723da5f860ef]  
[2024-05-15 08:50:42,372: INFO/ForkPoolWorker-2] Task create_storage_analytics_snapshot started
[2024-05-15 08:50:44,205: ERROR/ForkPoolWorker-2] Task analytics.tasks.create_storage_analytics_snapshot[fb0db7bd-5e81-4e1c-9946-723da5f860ef] raised unexpected: ClientError('An error occurred (404) when calling the HeadObject operation: Not Found')
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/celery/app/trace.py", line 382, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/celery/app/trace.py", line 641, in __protected_call__
    return self.run(*args, **kwargs)
  File "/app/src/apps/analytics/tasks.py", line 49, in create_storage_analytics_snapshot
    dataset.save()
  File "/app/src/apps/datasets/models.py", line 72, in save
    self.file_size = self.data_file.size / 1024
  File "/usr/local/lib/python3.8/site-packages/django/db/models/fields/files.py", line 69, in size
    return self.storage.size(self.name)
  File "/usr/local/lib/python3.8/site-packages/storages/backends/s3boto3.py", line 550, in size
    return self.bucket.Object(self._encode_name(name)).content_length
  File "/usr/local/lib/python3.8/site-packages/boto3/resources/factory.py", line 339, in property_loader
    self.load()
  File "/usr/local/lib/python3.8/site-packages/boto3/resources/factory.py", line 505, in do_action
    response = action(self, *args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/boto3/resources/action.py", line 83, in __call__
    response = getattr(parent.meta.client, operation_name)(**params)
  File "/usr/local/lib/python3.8/site-packages/botocore/client.py", line 357, in _api_call
    return self._make_api_call(operation_name, kwargs)
[...]

PR #1454, opened 2024-05-23
Related to issue #1451: Catch non-existant object error and clear object attribute in DB accordingly

OhMaley:
This piece of code avoids crash when trying to access the size of a file that does not exist anymore in the storage. If an error is caught it then clears the file_size and data_file attribute in the DB.

# Issues this PR resolves
Should solve Problem 1 of issue #1451 

# Simple test
A simple test would be to:
- create a private competition
- upload a submission
- change in the database (manually or by using the django admin interface) the `file_size` attribute to -1 of one of the submission's dataset (table datasets_data)
- remove in the storage (by using the minio interface) the corresponding file (its path is the value of the column `data_file` of the row in which you changed the `file_size` attribute) 
- Open up the shell_plus console inside the container of django
- Call the` save()` method of the dataset object you modified. It will do some checks on the existence of the file in the storage and do the necessary updates. It should be something in those lines:
```py
docker-compose exec django ./manage.py shell_plus
>>> d = Data.objects.get(id=<your_dataset_id>)
>>> d.save()
```
- Check that their is no crash and that some changes occured in the DB

# Use case test
- Remove a file from the storage
- Wait for the next create_storage_analytics_snapshot task to be ran (or trigger it manually)
- Check that everything worked correctly

# Checklist
- [x] Code review by me 
- [x] Hand tested by me (on a local setup)
- [x] I'm proud of my work
- [x] Code review by reviewer

### DIFF
--- a/src/apps/competitions/models.py
+++ b/src/apps/competitions/models.py
@@ -3,6 +3,7 @@ import uuid
 import os
 import io
 
+import botocore.exceptions
 from django.conf import settings
 from django.contrib.sites.models import Site
 from django.contrib.postgres.fields import JSONField
@@ -11,6 +12,7 @@ from django.db import models
 from django.db.models import Q
 from django.urls import reverse
 from django.utils.timezone import now
+from decimal import Decimal
 
 from celery_config import app, app_for_vhost
 from chahub.models import ChaHubSaveMixin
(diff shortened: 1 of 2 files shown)


Didayolo:
I merge this to a branch on codalab/codabench to trigger automatic testing.

Issue #1582, opened 2024-09-14
Release version


ihsaan-ullah:
We want to show release version on the website. Possible todos are the following:

1. Display release version in the footer
Solved by: #1583 
2. Make release clickable to take user to the release notes on GitHub
Solved by: #1583 
3. Make the release note dynamic (to be loaded from a configuration file
Solved by: #1592 
4. Make the release note automatic (to be updated in the configuration file automatically with each new release)
Solved by #1584

PR #1584, opened 2024-09-15
`Release PR#3` created workflow for new releases

ihsaan-ullah:
# @ mention of reviewers
@ObadaS 

Created a workflow and json file for release information.

# Issues this PR resolves
- #1582 -> Point 4

- [ ] We need to test if with a new release publish, this workflow is triggered

# Checklist
- [ ] Hand tested by me 
- [x] Ready to merge

### DIFF
--- a/.github/workflows/release-version-update.yml
+++ b/.github/workflows/release-version-update.yml
@@ -1,0 +1,36 @@
+name: Update version.json on release
+
+on:
+  release:
+    types: [published]  # Trigger when a new release is published
+
+jobs:
+  update-version:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
+    - name: Get release information
+      id: get_release
+      uses: dawidd6/action-get-latest-release@v3
+      with:
+        release: true  # Get details about the latest release
+
+    - name: Update version.json
+      run: |
+        echo '{
+          "tag_name": "${{ steps.get_release.outputs.tag_name }}",
+          "release_name": "${{ steps.get_release.outputs.name }}",
+          "published_at": "${{ steps.get_release.outputs.published_at }}",
+          "body": "${{ steps.get_release.outputs.body }}"
+        }' > version.json
+
+    - name: Commit version.json
+      run: |
+        git config --global user.name "github-actions[bot]"
+        git config --global user.email "github-actions[bot]@users.noreply.github.com"
+        git add version.json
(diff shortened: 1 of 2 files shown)


ObadaS:
Looks good on my end, we only need to test it now

1. I see in the code that you are using `>`, what do you think about using `>>` to have a local file in the repo with all the changes ?
2.  The code doesn't specify a branch, does it push the develop or master branch ? I think we want it to push on the master branch, otherwise we would need to do a separate PR to bring the updated file into master after we already merged into master (since we only publish the Release notes once we have successfully merged into master)

The CircleCI test fails with the same error I have on my local machine when trying to deploy the current develop branch. Seems like there is a problem with the Python version used, which could be an easy fix.

ihsaan-ullah:
> Looks good on my end, we only need to test it now
> 
> 1. I see in the code that you are using `>`, what do you think about using `>>` to have a local file in the repo with all the changes ?

`>` will rewrite the file while `>>` will append. I think we want to rewrite the file


> 2. The code doesn't specify a branch, does it push the develop or master branch ? I think we want it to push on the master branch, otherwise we would need to do a separate PR to bring the updated file into master after we already merged into master (since we only publish the Release notes once we have successfully merged into master)

I have updated the workflow to push changes to `develop` branch and automatically create a PR. We should test it on github if this works or not because I have never worked on workflows. I chose develop branch so that everything is consistent and there is no mismatch between develop and master

Didayolo:
Here, `version.json` looks like this:

```json
{
    "tag_name": "1.1.0",
    "release_name": "Second release",
    "published_at": "November 09, 2023",
    "body": ""
}

However, on the PR#2 (#1592) that'll be merged to develop, we have:

```json
{
  "tag_name": "v1.11.0",
  "release_name": "Release 1.11.0",
  "published_at": "2024-09-16",
  "body": "",
  "release_url": "https://github.com/codalab/codabench/releases/tag/v1.11.0"
}
```

That looks problematic

ihsaan-ullah:
json updated in the latest commit

Didayolo:
So should we merge this to develop? @ihsaan-ullah @ObadaS

ihsaan-ullah:
> So should we merge this to develop? @ihsaan-ullah @ObadaS

@ObadaS do you know when the workflow will be triggered? if this is in develop or master?

ObadaS:
It might not get triggered because the workflow file does not exist yet, and there is nothing that was published after the merge happens. 
But it should trigger automatically, after the merge, when I publish the 1.12 release

PR #1576, opened 2024-09-03
Merge develop into master

Didayolo:
# Testing

Deployed at https://codabench-test.lri.fr/

# Manual intervention

1. Copy and paste last storage_inconsistency files from /logs folder to /var/logs/ folder. Mind the 's' at the end of 'logs' (#1575).

```
cd codabench
cp -r logs/* var/logs
```

2. Re-build docker containers (#1593).

```
docker compose down
docker system prune -af
docker compose up -d
```

# Changes

- #1522 
- #1565
- #1562
- #1567
- #1560
- #1568
- #1569
- #1571
- #1575, #1609
- #1588
- #1583
- #1592 
- #1584
- #1593
- #1626
- #1613
- #1614
- #1612 
- #1630
- #1631
- #1634
- #1581

PR #1597, opened 2024-09-26
convert orphan files deletion task to async using celery task. Update celery config to make it possible

OhMaley:
Make the orphan files deletion task asynchronous.
This comes with few changes to make it possible:
- Enable the backend results of the celery tasks. This allows to store the states of any celery tasks. We use this to check is the task is running or not
- upgrade the celery package to the closest version that fixes a known bug on incompatible version with python 3.9 that manifests only when working with celery task results
- Fixes the patch number of the python version in the Dockerfile to match the one in the pyproject.toml file (else poetry won't be happy)
- Add an API route to check the status of the orphan files deletion task
- Update the frontend to reflect the status of the deletion task

# For testing
Do not forget to build the docker images !

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work


Didayolo:
@OhMaley Does this fix the bugs reported in #1576? Or it is just an addicional enhancement?

OhMaley:
Yes it avoids the server being stuck when running the deletion

PR #1738, opened 2025-01-31
Submission soft-delete functionality added

ihsaan-ullah:
# @ mention of reviewers
@adiprerepa 

...

# Issues this PR resolves
...

- [ ] add checklist here

[link]('#') to any relevant files (or drag and drop into github)

# Checklist
- [ ] Code review by me 
- [ ] Hand tested by me 
- [ ] I'm proud of my work
- [ ] Code review by reviewer
- [ ] Hand tested by reviewer
- [ ] Ready to merge


Didayolo:
@ihsaan-ullah 

I think the deletion button should only appear in the resources interface:

Capture d’écran 2025-02-04 à 16 26 07


and not in the submissions table from the competition:

Capture d’écran 2025-02-04 à 16 26 16


By the way, it looks like only new submissions (after deployment of this feature) are possible to delete:

Capture d’écran 2025-02-04 à 16 28 38

Didayolo:
OK, works for me. I think the deletion from resources interface is not working then, only from the competition page.

Didayolo:
- Updating the error message on resources interface (you can go to the competition etc.)
- Testing big files to see if you save the quota

ihsaan-ullah:
@ObadaS this is interesting. I will check on my side with a big file to see why full space is not freed.

ihsaan-ullah:
@ObadaS I checked on my side and I think when a submission is processed, some additional files are stored somewhere that takes additional space. Below are the details of my test

- No file upload
    Quota: 0.0B / 15.0GiB
- Uploaded Iris bundle:
    Quota: 907.1KiB / 15.0GiB
- Uploaded big submission (zip size: 1.34 GB)
    Quota: 1.3GiB / 15.0GiB     --- (1.40 GB)
- Uploaded big submission (zip size: 1.34 GB)
    Quota: 2.5GiB / 15.0GiB     --- (2.68 GB)
- When both submissions ran successfully
    Quota: 5.0GiB / 15.0GiB
- Deleted submission 1
    Quota: 3.8GiB / 15.0GiB
- Deleted submission 2
    Quota: 2.5GiB / 15.0GiB

ObadaS:
@ihsaan-ullah I think you can get more information about the files that are stored and created by looking at the logs of the compute_worker. It seems like, after downloading the file and doing the work needed, it generates a new file that gets uploaded to the storage. If I remember well, I read something about predictions files, since the ingestion and scoring program can be different, the generated files needs to be uploaded to the storage ?
I will look into it later and try to get logs about those mysterious files.

I agree about changing the size unit to GB to be less confusing since people are more used to this. If not too complicated, it would also be nice if the size change dynamically (shows in MB or KB if needed, instead of showing 0GB or 0.1GB)

ihsaan-ullah:
> I agree about changing the size unit to GB to be less confusing since people are more used to this. If not too complicated, it would also be nice if the size change dynamically (shows in MB or KB if needed, instead of showing 0GB or 0.1GB)

Sounds good. I will check this

ihsaan-ullah:
@ObadaS @Didayolo 

### I made the following updates:
- size formatter functions in both python and js updated to calculate size in GB/MB/KB instead of GiB/MiB/KiB 
- updated soft-deletion to fix the problem with remaining space that was left even after deleting everything
- changed default user quota size from 15 GiB to 15 GB
- updated user quota calculation to use GB/MB/KB instead of GiB/MiB/KiB 

### Some problems/thoughts
- In the calculation of storage analytics I found the use of GiB/MiB/KiB. Maybe there are more places where we have such inconsistency. I suggest that we convert this point into a separate issue and fix it later.
- I found redundant fields: `prediction_result_file_size`, `scoring_result_file_size`, `detailed_result_file_size`. These only keep the size of the files while the file fileds are also present and we can take the size from the file fields directly. We can ignore this point or we clean this up later in a separate issue/PR

Didayolo:
@ihsaan-ullah I confirm that now deleting submission does decrease the quota accordingly.

I've just seen something a bit shady.

I uploaded a submission with a size of 442 MB: 

Capture d’écran 2025-02-10 à 16 56 29

When the submission was being processed, I could see a consistent quota:

Capture d’écran 2025-02-10 à 16 56 42

But at the end of the computation, the quota used by the submission doubled:

Capture d’écran 2025-02-10 à 16 56 24

Is that normal?

ihsaan-ullah:
> @ihsaan-ullah I confirm that now deleting submission does decrease the quota accordingly.
> 
> I've just seen something a bit shady.
> 
> I uploaded a submission with a size of 442 MB:
> 
> Capture d’écran 2025-02-10 à 16 56 29
> When the submission was being processed, I could see a consistent quota:
> 
> Capture d’écran 2025-02-10 à 16 56 42
> But at the end of the computation, the quota used by the submission doubled:
> 
> Capture d’écran 2025-02-10 à 16 56 24
> Is that normal?

This needs a bit investigation but I think this comes from the prediction files stored after the submission is processed. If you remember, the prediction zip also includes the submission itself. BUT we should look into it together with my other points in my last comment

Issue #1766, opened 2025-03-04
🔥 Whitelist emails are leaked from API & should be case-insensitive


hzxie:
# Todos:

- [ ] Use whitelist emails in lower case
- [ ] Convert user email to lower case on registration to avoid multiple accounts creation with the same email

# Problem
I set the whitelist emails for a competition on CodaBench. However, I noticed two issues:

- **Security Issue:** The whitelist emails are exposed via the API (https://www.codabench.org/api/competitions/COMPETITION_ID/), which could be a privacy concern. These emails should not be publicly accessible. This issue persists despite the closure of #1366.
- **Case-Sensitivity Issue:** Some users registered with ALL CAPITALIZED email addresses, while the whitelist emails were stored in all lowercase. As a result, they were not recognized as whitelisted participants.

**Expected Behavior**

- Whitelist emails should be hidden or properly secured in the API.
- Email matching should be case-insensitive so that users can join the competition regardless of how they entered their email address.

**Steps to Reproduce**

1. Set up a competition with whitelist emails in all lowercase.
2. A user registers with the same email but in all uppercase (or mixed case).
3. The user is not recognized as a whitelisted participant.
4. Check the API and notice that the whitelist emails are exposed.

hzxie:
This screenshot illustrates the email exposure through the API.

![Screenshot](https://github.com/user-attachments/assets/08a248d3-bfe8-44c2-b067-32e0aff64324)

ihsaan-ullah:
Hi @hzxie,

You see whitelist emails because you are an organizer of this competition. Can you check from another account or fetch a competition that is not yours.

ihsaan-ullah:
Users with same emails but with lower and uppercases can register multiple accounts on codabench e.g. I can use the same email `ihsaN@gmail.com` and `ihsan@gmail.com` to create two accounts. I think we should not allow this and store the registered user's email in lowercase.

ihsaan-ullah:
I have added some todos to the first comment

hzxie:
> Users with same emails but with lower and uppercases can register multiple accounts on codabench e.g. I can use the same email `ihsaN@gmail.com` and `ihsan@gmail.com` to create two accounts. I think we should not allow this and store the registered user's email in lowercase.

Emails are not case-sensitive. For instance, AAA@gmail.com and aaa@gmail.com will be delivered to the same inbox.

hzxie:
[...]

PR #1769, opened 2025-03-05
Email in lowercase

ihsaan-ullah:
@Didayolo 

# Description
- On signup emails are stored as lowercased to not allow users to create multiple accounts using the same email but in different cases
- White list emails are now checked in lower case to avoid any inconsistencies  

# Issues this PR resolves
- #1766

- [x] signup with a dummy uppercased email e.g. `AAA@gmail.com` and check in your profile that it is stored as `aaa@gmail.com`
- [x] Add `AAA@gmail.com` to a competition whitelist and check that when you join the competition with `aaa@gmail.com` it approves you automatically

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/src/apps/api/views/competitions.py
+++ b/src/apps/api/views/competitions.py
@@ -304,7 +304,8 @@ class CompetitionViewSet(ModelViewSet):
             data.pop('whitelist_emails', None)
             # Loop over whitelist emails and add them back to whitelist emails in dict format
             for email in whitelist_emails:
-                data.setdefault('whitelist_emails', []).append({'email': email})
+                # user lower case email because some emails in the whitelist may have upper case letters 
+                data.setdefault('whitelist_emails', []).append({'email': email.lower()})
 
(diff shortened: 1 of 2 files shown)


hzxie:
Thanks for creating this PR.

Didayolo:
@ihsaan-ullah 

For me this change can be slightly problematic: Someone registering with an uppercase in the email will not be able to login typing the email in the same way the next time. 

I think we should either:
- Prevent from filling the form with uppercase, 
- or allowing to login with uppercase in the field (even if we sanitize it before comparing it to database). Maybe this method is prefered.

ihsaan-ullah:
> * or allowing to login with uppercase in the field (even if we sanitize it before comparing it to database). Maybe this method is preferred.

I updated the code to use lowercased email during login. This will allow users to use both lower and upper cased emails

Issue #1780, opened 2025-03-12
Issue to log in


nicolas124113:
# TODO:
- [x] Do not allow speical characters in username on signup

# Problem description
Hi, I just created an account on Codabench, I received the verification email but I get an error every time I try to log in, I typed my username with a tilde, I don't know if that might have caused it somehow or if it has something to do with it. Could I delete the account and create another one with the same email (as I describe it is not possible, I have also tried from the mobile and I get the same error)?

ihsaan-ullah:
Can you check if you can login using email and password?

ihsaan-ullah:
@ObadaS can we check logs for this issue. This seems a bit strange

Jorge1947890:
Same problem here, I used an accent (tilde) in my username and I get the same error. What can I do?

ObadaS:
@Jorge1947890 Can you give me your codabench username so that I can change it to Jorge1947890 (or you can also give me what new username you want me to change it to)

nicolas124113:
> Hello [@nicolas124113](https://github.com/nicolas124113), I have fixed your issue. The problem was coming from the special accent in your name. For now, I have changed your codabench username to nicolas124113 while we figure out why having accents in the username makes it not work. We will keep you updated and I will change your username back to Nicolás after the issue is fixed in the code.
> 
> Clear your cache to login again

@ObadaS thank you for solving it

Jorge1947890:
> [@Jorge1947890](https://github.com/Jorge1947890) Can you give me your codabench username so that I can change it to Jorge1947890 (or you can also give me what new username you want me to change it to)

Sure, my username is jorgegutiérrez and I want to change it to jorgeguti194

ObadaS:
@Jorge1947890 I have changed your username to jorgeguti194, you should be able to log in now (don't forget to clear your cache first)

monica570:
I signed up with an accent in my username (mónica) and I get the same error. How can I change it?

monica570:
New: monipazos

ObadaS:
It is done, you should be able to log in now

PR #1784, opened 2025-03-14
Do not allow special chars in usernames

ihsaan-ullah:
@Didayolo @ObadaS 

Special characters in username are problematic when django creates URLs. This PR restricts users to not use special characters in their usernames during signup

# Issues this PR resolves
- #1780 

- [x] Check that you cannot create an account if you have special character, capital letter in your username

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/src/apps/profiles/forms.py
+++ b/src/apps/profiles/forms.py
@@ -1,3 +1,4 @@
+import re
 from django import forms
 from django.contrib.auth.forms import UserCreationForm
 from .models import User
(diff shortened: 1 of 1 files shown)


ObadaS:
Wouldn't it be better, if possible, to allow special character in the name but change the way Django creates URL to "sanitize" the input first, removing the special characters ?

With this fix, people that have special characters in their name won't be to create an account without having to remove them.

ihsaan-ullah:
I think special characters should not be allowed in usernames. This was the case in the code but it was not working. In this PR I have changed the code to fix that code.

PR #1799, opened 2025-03-26
Maintenance mode option added in Caddyfile, used by creating a maintenance.on file in the maintenance_mode/ directory

ObadaS:
Added a maintenance mode to Codabench using Caddy and a maintenance.on file that is created (and deleted) when needed.

- [x] Create the `maintenance.on` file in the new maintenance_mode folder then check the homepage
- [x] Delete the `maintenance.on` file to make sure the homepage goes back to normal

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge


ObadaS:
@Didayolo We can use this to deploy the next update instead of shutting down Caddy. At least like this, people can see that the site is on maintenance instead of just having a generic "Bad Gateway" message

Didayolo:
@ObadaS We need to document this in the Wiki

ObadaS:
Yes, I will update the wiki page for Administrators to explain this feature and show how to use it.

ObadaS:
@Didayolo I have fixed the problem. 
`try_files` was not allowing any other files to be loaded (so offline.png was not usable). I changed the caddy directive to `redir` (a redirect) to allow other files to be loaded at the same time as the maintenance.html file.

PR #1830, opened 2025-04-16
Clarify warning message "Detailed results not written" in compute_worker.py

Didayolo:
# Description

Update the warning message to avoid confusion.

# Checklist
- [ ] Code review by reviewer
- [ ] Ready to merge


Didayolo:
If we're going to update the compute worker, we should use this occasion to solve the "Stuck in scoring when compute worker" crashes at the same time

PR #1838, opened 2025-04-25
Add hide_score_output option

Didayolo:
# Description

New phase option "Hide score output" (also a `competition.yaml` option):

Capture d’écran 2025-04-25 à 12 18 28

If checked, it prevents from downloading "Output from scoring step":

Capture d’écran 2025-04-25 à 12 20 41

It is useful, because the current option "Hide submission output" hide all logs and is too extreme. But in some cases, typically with hidden leaderboard columns, we don't want participants to be able to read the `scores.txt` file.

- [x] Edit phase setting to set "Hide score output" to True
- [x] Check that it prevents from downloading the scores.txt file (expect as an admin)
- [x] Try to play with the two settings "Hide submission output" and "Hide score output"
- [x] If it is set to False, the download should be allowed again
- [x] Try the `competition.yaml` option `hide_score_output: True` (it is a phase option)

# To check

I added 3 migration files. Maybe we can avoid having that many.

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Documentation updated
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge


Didayolo:
@ObadaS Would you be able to review this one?

ObadaS:
Yes I can review it. 
Quick question for this part `Check that it prevents from downloading the scores.txt file (expect as an admin)`, does it also include the competition organizer or only the plateform admins ?

ObadaS:
` Try the competition.yaml option hide_score_output: True (it is a phase option)` is not working. I modified the Iris competition (changed the phase name and added the option in the `competition.yml` file) but the option does not get turned on by default once I upload the bundle to create the competition.

Here is the bundle I am using (I modified the final phase)
[iris3.zip](https://github.com/user-attachments/files/19907744/iris3.zip)

After uploading, I tried to submit something on an another account but I could still download the output. I then checked the competition settings from the interface and I saw that the option is turned off.

Didayolo:
> ` Try the competition.yaml option hide_score_output: True (it is a phase option)` is not working. I modified the Iris competition (changed the phase name and added the option in the `competition.yml` file) but the option does not get turned on by default once I upload the bundle to create the competition.
> 
> Here is the bundle I am using (I modified the final phase) [iris3.zip](https://github.com/user-attachments/files/19907744/iris3.zip)
> 
> After uploading, I tried to submit something on an another account but I could still download the output. I then checked the competition settings from the interface and I saw that the option is turned off.

En effet, je n'ai mis à jour que l'unpacker "v2", je n'ai pas pensé au cas de figure où on edit un bundle CodaLab pour y ajouter cette nouvelle option.

Didayolo:
@ObadaS Done!

ObadaS:
The option now works correctly for the custom competition I modified.

PR #1851, opened 2025-05-14
Add hide_prediction_output option

Didayolo:
# Description

New phase option "Hide prediction output" (also a `competition.yaml` option).

If checked, it prevents from downloading "Output from prediction step".

Basically the same as #1838 (see screenshots there) but with prediction output.

It is useful especially when data are confidential.

- [x] Edit phase setting to set "Hide prediction output" to True
- [x] Check that it prevents from downloading the prediction file (expect as an admin)
- [x] Try to play also with the settings "Hide submission output" and "Hide score output"
- [x] If it is set to False, the download should be allowed again
- [x] Try the `competition.yaml` option `hide_prediction_output: True` (it is a phase option)

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Documentation updated
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge


ObadaS:
Isn't it supposed to be `hide_prediction_output: True` instead of `hide_score_output: True` ?

I tested it with `hide_prediction_output: True`  and it works fine.

Issue #1881, opened 2025-06-18
Do not allow emails with `*`


ihsaan-ullah:
In signup do not allow users to register with emails with `*` in it.

PR #1882, opened 2025-06-18
Do not allow signup with email with `*`

ihsaan-ullah:
@Didayolo 

# Description
Codabench was allowing users to use emails with `*` e.g. ihsan*@gmail.com was considered a valid email. Now this will raise an error.

Screenshot 2025-06-18 at 4 38 44 PM

# Issues this PR resolves
- #1881

- [x] try signing up with an email with `*` in it

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/compute_worker/compute_worker.py
+++ b/compute_worker/compute_worker.py
@@ -588,7 +588,7 @@ class Run:
         Function responsible for running program directory
 
         Args:
-            - program_dir : can be either ingestion program or program(submission or scoring)
+            - program_dir : can be either ingestion program or program/submission
             - kind : either `program` or `ingestion`
         """
         # If the directory doesn't even exist, move on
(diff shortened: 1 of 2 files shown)


Didayolo:
@ihsaan-ullah 

Shouldn't we make this more general than just not allowing `*` ?

Like for usernames, I think more characters should be disallowed.

```python
    def clean_username(self):
        data = self.cleaned_data["username"]

        # Check if username has allowed characters only
        # Allow only lowercase letters, numbers, hyphens, and underscores
        if not re.match(r"^[a-z0-9_-]+$", data):
            raise forms.ValidationError("Username can only contain lowercase letters, numbers, hyphens, and underscores.")

        # Check username length
        if (len(data) > 15) or (len(data) < 5):
            raise forms.ValidationError(
                "Username must have at least 5 characters and at most 15 characters"
            )
        return data

    def clean_email(self):
        email = self.cleaned_data["email"]
        if "*" in email:
            raise forms.ValidationError("Email address cannot contain the '*' character.")
        return email
```

PR #1964, opened 2025-08-08
Fixed MC image version and bumped MinIO version

### LINKED ISSUE #1957: lock down image version of minio/mc
like with the minio image, the version of the minio/mc image should probably be set to a specific version in the docker-compose.yml

this should also prevent an issue like #1926 suddenly appearing

ObadaS:
As recommended here #1957 , I locked down the version of the MC image and updated the MinIO image at the same time

# Issues this PR resolves
#1957

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -53,7 +53,7 @@ services:
   # Minio local storage helper
   #-----------------------------------------------
   minio:
-    image: minio/minio:RELEASE.2020-10-03T02-19-42Z
+    image: minio/minio:RELEASE.2025-04-22T22-12-26Z
     command: server /export
     volumes:
       - ./var/minio:/export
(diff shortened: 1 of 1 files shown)


ObadaS:
I think 'mc mirror' might be the only option we have, I read some other comments on it that says the same : https://github.com/minio/minio/discussions/15967. 

Here is the [official guide](https://docs.min.io/community/minio-object-store/operations/deployments/baremetal-migrate-fs-gateway.html) to migrate from the old FS to the new one. I think we are migrating FS not Gateway, so users should be choosing the FS migration steps when following the guide

PR #1991, opened 2025-08-28
Merge develop into master (1.21.0)

Didayolo:
# Test

To be deployed at https://codabench-test.lri.fr/

# Manual intervention

- #1963
TODO

# Changes

TODO


ObadaS:
This concerns https://github.com/codalab/codabench/pull/1955

@ihsaan-ullah 
Hello, I update the develop branch on the test server, rebuild all docker images, then ran `docker compose exec django ./manage.py migrate`, however, when I upload a competition, I can not see it in the new menu. I tried multiple time on my local instance and on the test server, with different competitions. The competitions are created and can be accessed if you know the link, but it is not showing in the new page, even when pressing `Organizing` checkbox
image


Also, we are losing some functionality with the new page, since we can't publish/unpublish, edit or delete competitions quickly from the page. Do you think it could be added back ?
image

If you are not sure where to add it, I think the way Codalab does it is a good start.
image

ObadaS:
- Go to https://codabench-test.lri.fr/competitions/public/?page=1
- Create a new competition by uploading a bundle
- Go to back to https://codabench-test.lri.fr/competitions/public/?page=1 and look for the competition that you just created. You won't be able to find it because it is not public

After reading your comment, I tried to publish a competition I uploaded, and it appeared on the page. How can I see all the private competitions that I have uploaded ?

ihsaan-ullah:
We can always improve the flow/UI if there is any confusion

Didayolo:
> We can always improve the flow/UI if there is any confusion

It is just a matter of getting used to it. Maybe we could communicate about the change.

Personally I love this new interface, generally it's lighter and more intuitive.

Issue #1641, opened 2024-10-28
Submissions that throw exceptions are marked as completed instead of failed


ndido98:
I am hosting a private competition on Codabench (https://www.codabench.org/competitions/4188/) and we are running into an issue when the participants’ submissions fail the scoring step.

In particular, when the scoring program fails (mostly because due to an error from the participant in formatting the submission package), the submission should be marked as failed and not count towards the submission limit per day/person. However, it looks like the exit code of the scoring program is not checked and all submissions are marked as finished instead.

Didayolo:
Hi @ndido98, how does your scoring program handle the formatting error?

To trigger the submission failure, the scoring program should raise an error.

ndido98:
Hi @Didayolo, currently the scoring program just raises an exception, and as a consequence it returns an exit code of 1. The scores.txt file is created but it's empty.

Didayolo:
Indeed, I am able to see the scoring error logs. I'm surprised the submissions are marked as finished. 

I'm actually not sure what triggers the "Failed" status. Maybe it is because the `scores.txt` file exist? This requires a little investigation.

ndido98:
From what I gathered by taking a look at the source code I found that the exit code of the scoring program is never checked:
https://github.com/codalab/codabench/blob/7cf7a6ed8e091bf1160933f3dd360c31ca704d43/compute_worker/compute_worker.py#L878-L881

Probably a submission should be marked as failed if either ingestion (if present) or scoring programs have a non-zero exit code.
The reason it fails when no `scores.tx` file is present is because this exception is triggered:
https://github.com/codalab/codabench/blob/7cf7a6ed8e091bf1160933f3dd360c31ca704d43/compute_worker/compute_worker.py#L903

Not outputting a `scores.txt` file (thus triggering the exception) could be a workaround for the time being, but still I believe that checking the exit code of the process should be preferred.

Didayolo:
Interesting, that may explain the "Submission stuck in Scoring status" when compute worker crashes problem, reported in #1205.

Issue #1205, opened 2023-10-31
Compute worker issues


Didayolo:
- [ ] Docker pull error during scoring are written in ingestion stderr instead of scoring stdrerr
- #1204

- [ ] Submissions stuck in "Scoring" state instead of "Failed" when the compute worker crashes
- #1203
- #1184

Didayolo:
Docker pull failing: a worker was ill, we removed it from the queue.

gorkamunoz:
Hello, I am recently also facing the Docker pull fail (in particular `Pull for image: codalab/codalab-legacy:gpu returned a non-zero exit code! Check if the docker image exists on docker hub. Command '['docker', 'pull', 'codalab/codalab-legacy:gpu']' returned non-zero exit status 1.` ).

Is there an existing solution already? Thanks (and thanks for the awesome plaftorm!)

Didayolo:
Hello @gorkamunoz,

Do you have your own compute workers? When you try to run manually the `docker pull codalab/codalab-legacy:gpu` command there, what happens?

This may be related to your configuration or network and not directly to our code base.

> Thanks (and thanks for the awesome plaftorm!)

Thank you for the kind words!

Didayolo:
> Hi, thanks to for the quick response. I am not using my own compute workers, hence I understand this is running on your side?

It depends on the competition. Are you a competition organizer or participant?

By default the competitions run on our workers, but organizers can also create their own queue to receive the jobs (the submissions) on their machines. It looks like the competition in question is using a `:gpu` docker image, so probably it is not supposed to run on our servers (the default queue does not feature GPU).

Didayolo:
Thank you for the information. We'll investigate our compute workers to resolve this inconsistency.

PR #2030, opened 2025-10-11
Worker status to FAILED instead of SCORING or FINISHED in case of failure

Didayolo:
# Description

Sometimes failure in the scoring program did not update the status in the right way.
This PR should improve this process.

I also took the chance to improve the logging, typically avoiding the `"Error" + e` format and replacing it by `f"Error: {e}"`, making the code more robust.

Moved the ingestion and scoring hostname update code from the star() method to the prepare() method, so that we can get the hostname earlier (allowing us to know which worker was handling a submission if it fails during the preparation phase for example)

# Issues this PR resolves
- First point of #1205 ?
- #1641

# Installation

- Need to **re-build compute worker**

- [x] Check that submissions (results and code) are working fine
- [x] Check that failing submissions get the status `FAILED`

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/compute_worker/compute_worker.py
+++ b/compute_worker/compute_worker.py
@@ -452,6 +452,7 @@ class Run:
             logger.error(f"This error might result in a Execution Time Exceeded error: {e}")
             if os.environ.get("LOG_LEVEL", "info").lower() == "debug":
                 logger.exception(e)
+            raise SubmissionException("Could not connect to instance to update detailed result")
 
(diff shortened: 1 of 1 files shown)


Didayolo:
@ObadaS This PR is now ready for review. Not a big change but still an improvement of the compute worker.

PR #2055, opened 2025-11-06
Deploy codabench using kubernetes

### LINKED ISSUE #1933: Running Codabench on Kubernetes
Hi,

thanks a lot for the great documentation on how to setup a codabench instance. @pauloguilhermepp and I are trying to run Codabench on Kubernetes and we were wondering if you have any recommendations or experience with that? We have been working on a helm chart, that @pauloguilhermepp would be happy to contribute if there is any interest from your side.

Our approach at the moment is to replicate the environment from the docker compose that you provide to this helm chart and we were wondering how to approach the execution of user submissions. If we understand correctly, the compute worker interacts with the docker daemon on the host via the mounted socket.  There could be ways of doing this in Kubernetes as well, e.g. via interacting with the Container Runtime Interfaces on the nodes but we were wondering if an approach using the Kubernetes API itself would be helpful. In this setting, the django component (or another component) would interact with the Kubernetes API server to run submission pods which Kubernetes would then schedule and orchestrate. 

We are happy to hear your opinions on that!

Best,
Paulo and Hannes

pauloguilhermepp:
#  Deploy Codabench Using Kubernetes
This PR aims to make Codabench deployable using Kubernetes.

# Main Changes
To deploy Codabench using Kubernetes, we:
* Added Helm charts for all core components.
* Updated Dockerfiles to improve compatibility with Kubernetes deployments.

# Issues this PR resolves
* Closes #1933

# How to Test it
1. Prerequisites
    * [Kubectl](https://kubernetes.io/docs/reference/kubectl/)
    * [Helm](https://helm.sh/)
    * Access to a kubernetes cluster (like [minikube](https://minikube.sigs.k8s.io/docs/start/?arch=%2Flinux%2Fx86-64%2Fstable%2Fbinary+download) or [kind](https://kind.sigs.k8s.io/))
2. Start your cluster (eg, `minikube start`)
3. Fill in `values.yaml` properly
4. Deploy the Helm chart
```
helm install <project_name> . -n <namespace> --create-namespace -f <path_to_values>
```
5. Check the status of your pods
```
kubectl get pods
```

# Observations

* Tests mainly made at [1.19](https://github.com/codalab/codabench/tree/1.19.0). 

* To expose the codabench UI, you will have to install an [ingress controller](https://kubernetes.io/docs/concepts/services-networking/ingress/) and create an Ingress.

* This PR does not include a template for Minio yet. For now, testing requires an external S3 instance. We can also add the Minio chart if needed.

* The helm chart passes secret values directly via environment variables, which is not the best practice, as it will make them visible in deployment tools like ArgoCD. If you want, we can update this PR (or open a future one) with this change.

* Inside values.yaml, there is a section (env) with environment variables that are currently passed to multiple pods. This is similar to the .env setup in docker-compose, but as we sometimes had a hard time debugging which variable is used where, we would like to separate those in a later PR.

* This PR is making Codabench deployable using Kubernetes, but it is not changing how the compute-worker is running the user submissions. We plan to open another PR in the future, updating the compute-worker to run submissions using Kubernetes. We plan to make it configurable and using docker as default so that it will not break any current deployments. Please note that for the current way the compute worker works, the Kubernetes Pods need to mount a volume from the host node, to mount the Docker socket from the node, and run privileged Docker containers, which is not the best practice. This will mess with the storage on the node and will mess with the scheduling, as containers are directly created, bypassing the Kubernetes scheduler. Therefore, we will open another PR.

* The PR also contains the Dockerfiles that we are using. Ideally, we should see if they are still compatible with deploying Codabench using the original docker-compose setup and keep a single Dockerfile for each component. The images were updated because, ideally, we don't want to mount the code from a PersistentVolume or the host node but rather have the image contain the code.

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/charts/templates/django-deployment.yaml
+++ b/charts/templates/django-deployment.yaml
@@ -20,10 +20,11 @@ spec:
             - bash
             - -c
             - >
-              python manage.py collectstatic --noinput &&
+              python manage.py migrate --no-input &&
+              python manage.py collectstatic --no-input &&
               cd {{ .Values.django.workingDir }} &&
               watchmedo auto-restart -p '*.py' --recursive --
-              gunicorn asgi:application -w {{ .Values.django.gunicorn.workers }} -k uvicorn.workers.UvicornWorker -b :{{ .Values.django.port }} --capture-output --log-level {{ .Values.django.gunicorn.logLevel }}
+              python3 ./gunicorn_run.py
           env:
           {{- range $key, $value := .Values.env }}
             - name: {{ $key }}
--- a/charts/values.yaml
+++ b/charts/values.yaml
@@ -23,9 +23,6 @@ django:
   replicas: 1
   port: 8000
   workingDir: /app/src
-  gunicorn:
-    workers: 2
-    logLevel: info
 
 flower:
   image:
@@ -130,6 +127,8 @@ db:
   port: <database-port>
 
 env:
+  LOG_LEVEL: "info"
+  
   SECRET_KEY: <your-django-secret-key>
 
   DJANGO_SETTINGS_MODULE: settings.develop


charts/Chart.yaml
```diff
@@ -0,0 +1,34 @@
...
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "1.16.0"
+
+dependencies:
+  - name: rabbitmq
+    version: "14.7.0"
+    repository: "oci://registry.cern.ch/kubeflow/charts"
```
hahahannes: We will replace this one with the original upstream chart https://hub.docker.com/r/bitnamicharts/rabbitmq. Same for redis below.

charts/dockerfiles/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+
+# Set work directory before copying files
+WORKDIR /app
+
+# Copy only dependency descriptors first (for caching)
+COPY pyproject.toml poetry.lock ./
+RUN poetry install
+
+# Copy the rest of the application code
+COPY . /app
```
hahahannes: Regarding the discussion about the Dockerfiles. We saw that the django component was missing the code (https://github.com/codalab/codabench/blob/develop/Dockerfile) also in the latest develop branch. Therefor we added node builder image here as well as the copying of the code.

pauloguilhermepp:
Hi everyone.

We rebased the branch and also updated the Django image so it is self-contained.

We successfully installed the Chart using this updated image.

pauloguilhermepp:
Hi everyone,

With the conflicts fixed in the PR, it should now be possible to merge it. 

[Containerfile](https://github.com/pauloguilhermepp/codabench/blob/develop/packaging/container/Containerfile) is the only change that also affects the docker-compose setup. As we mentioned, the images must be self-contained to be deployed using Kubernetes. This means that all code used by the image must be contained within it.

The new Containerfile combines the [old Containerfile](https://github.com/codalab/codabench/blob/develop/packaging/container/Containerfile) with [Container.builder](https://github.com/codalab/codabench/blob/develop/packaging/container/Containerfile.builder) in such a way that all Django dependencies will be present in the final image. If you consider that this is incompatible with the docker-compose deployment, we can add a new Containerfile with such changes. 

Please, feel free to share your feedback.

Didayolo:
Thank you for your work!

Post-it: we'll need to rebase and resolve conflicts again after merging #2223.

---

EDIT: I thought this PR modified `compute_worker.py`. It's all good now, and `fix-worker-logs` is rebased.

Issue #2022, opened 2025-10-06
Remove container engine from compute worker and user python library instead


wlln:
At the moment, we have 4 Containerfiles:

  - Docker CPU
  - Docker GPU
  - Podman CPU
  - Podman GPU

Following this logic, when we have k8s support, we'll end up with 6 Containerfiles.

I reckon that we could have one single Containerfile with Podman/Docker/K8S support using python libraries (python-podman, python-docker, for instance). This would ease the k8s integration in the future.

Didayolo:
Done #2139

PR #2065, opened 2025-11-14
`Compute Worker` Use docker-py  instead of the current subprocess way of doing things (for podman and docker)

ObadaS:
Changed the code to use the Docker and Podman python packages instead of using the current subprocess way of launching containers for the compute worker.

For Podman containers using GPU, we need podman 5.4 minimum : https://github.com/containers/podman/pull/25171

With these changes, we only need 1 compute worker image which will be able to communicate with the Docker and Podman socket, as well as use the available GPU on the host machine without having to add anything inside the image.
This results in a much smaller image (around 400 MB, down from 1 GB for the Docker image and 700 MB for the podman image)

# New option in .env
You can now select which GPU you want to assign to the compute worker if you have multiple.
You can get the list by running `nvidia-ctk cdi list`. You then copy the name of the GPU you want and add it in the `.env` like this : 
```ini
# Replace *name* with the name as shown in the previous command. For example : nvidia.com/gpu=0
GPU_DEVICE=*name*
```

By default, if `GPU_DEVICE` is not set, it will take the value of `nvidia.com/gpu=all`

## New Github Workflow
I also added workflows to automatically build images and push them to Dockerhub when the `Dockerfile.compute_worker` file and `compute_worker/` directory receives changes. The image will have the `test` tag when made from the develop branch, `release tag` and `latest` tag when made from the master branch and the `branchName` tag when made from a branch that changes the files.

 Since we can not delete images with the Docker command, we will need to clean up the image registry from time to time.
# Issues this PR resolves
https://github.com/codalab/codabench/issues/2022

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/.github/workflows/build_cw_image-branch.yml
+++ b/.github/workflows/build_cw_image-branch.yml
@@ -17,11 +17,11 @@ jobs:
       - name: Check out repository code
         uses: actions/checkout@v5
       - name: Build Image
-        run: docker build -t codalab/competitions-v2-compute-worker:${{ github.ref_name }} -f Dockerfile.compute_worker .
+        run: docker build -t codalab/codabench-compute-worker:${{ github.ref_name }} -f Dockerfile.compute_worker .
       - name: Login to Docker Hub
         uses: docker/login-action@v3
         with:
           username: ${{ vars.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
       - name: Push image with branch name as tag
-        run: docker push codalab/competitions-v2-compute-worker:${{ github.ref_name }}
+        run: docker push codalab/codabench-compute-worker:${{ github.ref_name }}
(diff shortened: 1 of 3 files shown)


compute_worker/compute_worker.py
```diff
@@ -508,33 +509,66 @@ async def _run_container_engine_cmd(self, container, kind):
...
-            containerLogs = client.logs(container, follow=True, stream=True)
-            for log in containerLogs:
-                logger.info(log)
-        except docker.errors.NotFound as e:
+            if os.environ.get("CONTAINER_ENGINE_EXECUTABLE").lower() == 'docker':
+                client.start(container=container.get('Id'))
+                conatinerLogsDemux = client.attach(container, demux=True, stream=True, logs=True)
+            elif os.environ.get("CONTAINER_ENGINE_EXECUTABLE").lower() == 'podman':
+                client.containers.start(container)
+                conatinerLogsDemux = client.containers.attach(container, stream=True, demux=True)
```
wlln: You wrote conatinerLogsDemux instead of containerLogsDemux

PR #2095, opened 2025-12-23
EMAIL - replaced `DEFAULT_FROM_EMAIL` by `SERVER_EMAIL`

ihsaan-ullah:
# Description
Replaced `DEFAULT_FROM_EMAIL` by `SERVER_EMAIL` for sending emails

# Issues this PR resolves
- #2082 

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] CircleCi tests are passing
- [x] Ready to merge

PR #2094, opened 2025-12-23
General - Added new files for Governance, Privacy and About

ihsaan-ullah:
# Description
In docs, we were pointing to codalab-competitions wiki for Governance, Pricvacy and About. Now we have dedicated files in this repo for that.

# Issues this PR resolves
- #2083 -> Point 2

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] CircleCi tests are passing
- [x] Ready to merge

PR #1959, opened 2025-08-03
Django to 4.2.0

bbearce:
# django_to_4.2.0
This will be incremental but this is the PR to push django up to 4.2.0 LTS

# @ mention of reviewers
...

# A brief description of the purpose of the changes contained in this PR.
Upgrading django to major version 4.2.0 from 3.0.0

# Issues this PR resolves
...

# A checklist for hand testing
- [ ] circleci tests only so far

# Checklist
- [x ] Code review by me 
- [x ] Hand tested by me 
- [x ] I'm proud of my work
- [ ] Code review by reviewer
- [ ] Hand tested by reviewer
- [ ] CircleCi tests are passing
- [ ] Ready to merge

PR #2125, opened 2026-01-28
Merge Develop into Master (v1.23)

ObadaS:
# Manual Interventions
## Django
The Django version upgrade will require all the containers to be rebuilt:
```sh
docker compose build --no-cache
docker compose up -d
```

Afterward, you will need to run some migrations: 
```sh
docker compose exec django python manage.py migrate
```
If it asks you to do a makemigration instead, you will have to run the following command:
```sh
docker compose exec django python manage.py makemigrations --merge
```
You might need to modify `/app/src/apps/datasets/migrations/0014_merge_20251212_0942.py`  to remove the `datagroup` changes
```python
# Generated by Django 4.2.23 on 2025-09-08 12:32

from django.db import migrations, models
import utils.data
import utils.storage


class Migration(migrations.Migration):

        dependencies = [
            ("datasets", "0010_auto_20250218_1100"),
        ]

        operations = [
            migrations.AlterField(
                model_name="data",
                name="data_file",
                field=models.FileField(
                    blank=True,
                    null=True,
                    storage=utils.storage.PrivateStorageClass(),
                    upload_to=utils.data.PathWrapper("dataset"),
                ),
            ),
            migrations.AlterField(
                model_name="data",
                name="id",
                field=models.BigAutoField(
                    auto_created=True, primary_key=True, serialize=False, verbose_name="ID"
                ),
            ),
        ]
```
Notice that there are only two migrations in this file instead of 3. Simply delete the last one (it tries to change `datagroups` even after `datagroups` was deleted, which is why it fails)

---

# Release notes

- #2124

## Improvements
- #1959 
- #2126
- #2098 
- #2095 
- #2138
- #2094
- #2158 
- #2001

## Documentation
- #2157

### DIFF
--- a/src/apps/api/views/submissions.py
+++ b/src/apps/api/views/submissions.py
@@ -14,11 +14,10 @@ from rest_framework.settings import api_settings
 from rest_framework.viewsets import ModelViewSet
 from rest_framework_csv import renderers
 from django.core.files.base import ContentFile
-from django.http import StreamingHttpResponse
 
 from profiles.models import Organization, Membership
 from tasks.models import Task
-from api.serializers.submissions import SubmissionCreationSerializer, SubmissionSerializer, SubmissionFilesSerializer
+from api.serializers.submissions import SubmissionCreationSerializer, SubmissionSerializer, SubmissionFilesSerializer, SubmissionDetailSerializer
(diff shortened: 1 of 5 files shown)


Didayolo:
Deployed on https://test.codabench.org/

PR #2225, opened 2026-03-02
Replace Caddy with Nginx

ObadaS:
- Replace Caddy with Nginx, allowing us to route every connection through it (like rabbit, minio, postgres) instead of only HTTP(S) connections
- Add anti-scrapping measures
- Add robots.txt file
- Closed all the external ports of the containers except Nginx. Everything must go through Nginx now.
- Made all containers (except Nginx, Django, Compute Worker and Builder) unable to access the outside world.
- Add rate limit

# Manual Intervention
There are some new environment variable that will be needed in the .env file
```
HTTPS=False # Or True
RATE_LIMIT=5
```

If your `DOMAIN_NAME` contains :  you will need to remove it.
For example, `localhost:80` becomes `localhost`

# Checklist
- [ ] Code review by me 
- [ ] Hand tested by me 
- [ ] I'm proud of my work
- [ ] Code review by reviewer
- [ ] Hand tested by reviewer
- [ ] Ready to merge

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,6 +31,7 @@ jobs:
       - run:
           name: "Setup: Create directories for MinIO (cannot be made by docker for some reason)"
           command: |
+            sudo echo "127.0.0.1 minio" >> /etc/hosts
             mkdir -p var/minio/public
             mkdir -p var/minio/private
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -89,6 +89,8 @@ jobs:
           path: tests/test-results
       - store_artifacts:
           path: dockerLogs/
+      - store_artifacts:
+          path: var/log/
 
 workflows:
   version: 2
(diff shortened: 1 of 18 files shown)


ckravit:
I'm pretty sure Caddy can do this. Why force nginx? Caddy works really well especially for the SSL cert renewals. Can also easily customize Caddy to restrict some of the admin pages (rabbit, django, etc) to certain subnets.

ObadaS:
> I'm pretty sure Caddy can do this. Why force nginx? Caddy works really well especially for the SSL cert renewals. Can also easily customize Caddy to restrict some of the admin pages (rabbit, django, etc) to certain subnets.

While Caddy can do some of the new features in this PR, the main problem is that Caddy can only handle HTTP(S) connections. We can not use it to re-route AMQP connections for Rabbit.

However, we are also discussing internally about removing Caddy from the docker compose since it's not really related to Codabench, allowing people to use whatever reverse proxy they want to use, and giving an example configuration of what could be used.

ckravit:
Given some of the customizations I've done, I may consider [caddy-l4](https://caddyserver.com/docs/modules/layer4) for that instead of trying to recreate everything I've customized in nginx. I don't think it would add too much overhead and would be a simpler transition.

PR #2232, opened 2026-03-05
Auto Generation of django secret key

IdirLISN:
# Description:
Generation of a random secret key for django instance when building the app. 
Write the value on .env. 
each docker compose up -d, entrypoint.sh will read the variable in .env, if it doesn't exist it will generate one random secret key.
If a user want to deploy his own codabench instance, it will be secured with a unique secret key.
Enables django to use a private secret key for each new instance of it. 

# Issues this PR resolves
Secret key was clear and was the same for all instances. It means all codabench instances can authenticate to each other.  

- [ ]  just deploy and start the app. If it works, it means django found the secret key randomly generated.

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [ ] Code review by reviewer
- [ ] Hand tested by reviewer
- [ ] Ready to merge

### DIFF
--- a/src/settings/base.py
+++ b/src/settings/base.py
@@ -5,7 +5,6 @@ from celery.schedules import crontab
 from celery import signals
 import dj_database_url
 from .logs_loguru import configure_logging
-from django.core.management.utils import get_random_secret_key
 
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 # Also add ../../apps to python path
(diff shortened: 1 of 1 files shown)


ObadaS:
I can't make the `SECRET_KEY` appear in the .env file after launching the django container.

It also seems like you forgot to remove the `SECRET_KEY` in the .env_circle file, which means that the PR is not being fully tested

Didayolo:
Is the version from #2267 working fine? If so we can close this one and focus on #2267.

ObadaS:
> Is the version from #2267 working fine? If so we can close this one and focus on #2267.

Yes, I tested it locally and everything worked fine

Didayolo:
OK, I close this one. If we change our mind we can always re-open it.

PR #2237, opened 2026-03-06
All compute_work to run with Python 3.12

cjh1:
Fix use of `loop` in `gather(...)` to allow with Python 3.12.

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge


cjh1:
> Hello, I had the same problem when I upgraded the python version in #2200 but I fixed it like this : https://github.com/codalab/codabench/pull/2200/changes#diff-4445d9d2e4988a27d815ccde4804def3d361ceba5104a1ef8c094495bdb06c0bR1201
> 
> Is your way of fixing it better ? In my limited testing, the compute worker worked fine, but I haven't done a long test to make sure that not explicitly closing the loop doesn't cause any problems

I think the intent of the original code was to ensure that each `gather` operation was performed in a new event loop. I assume this was because of issues with running async code inside a worker. Your change does doesn't this, it now doesn't use the new event loop, it uses what was already set, as its no longer pass into `gather`. My change set the new evetn loop. The close is just to ensure cleanup, the original code had a potential resource leak by not cleaning up the event loop each time.

PR #2245, opened 2026-03-09
Submission pagination backend fixed

IdirLISN:
# A brief description of the purpose of the changes contained in this PR.
Pagination display for submission in competition UI.
Using the existing file pagination.py inside of submissions viewset.

Dynamic pagination class added in pagination.py. 
Get 50 objects by default.
Users can choose to get 50, 100, 500 or all objects.
"all objects" option is hard limited by max_page_size = 1000 per page.

Submission UI now displays 50 objects per page and a panel to enable the user to change this value. 

# Issues this PR resolves
- Closes #2220 

# A checklist for hand testing
- [x] Go to a competition with a lot of submissions and test the display of submissions inside of the competition page and inside of the organizer panel. 

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] CircleCi tests are passing
- [x] Ready to merge

screenshot:
image

image

PR #2239, opened 2026-03-07
Fix auto-migration to next phase

Didayolo:
# Description

Previously, the phase migration returned `N/a` scores on the leaderboard. The problem was that the submission was actually run on a task from the previous phase.

This simple change fixes:
- Scores displayed on leaderboard
- Submission migrated run on the right task(s)

# Issues this PR resolves
- Point **2. c)** of #1540 

# A checklist for hand testing
- [x] Make a leaderboard submission
- [x] Set up the next phase to change soon and enable "auto-migrate to this phase"
- [x] Wait and check that the submission run and is displayed well on leaderboard

If the auto-migration is not triggered, you can enter inside Django container, shell_plus, and run:
```python
from competitions.tasks import manual_migration
manual_migration(<FIRST_PHASE_ID>)
```

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] CircleCi tests are passing
- [x] Ready to merge

PR #2228, opened 2026-03-04
Fix submission upload page display

Didayolo:
# Description

### 1. Fact sheet inside a scroller if too long (#2106)

Capture d’écran 2026-03-04 à 13 10 23

### 2. "Autoscroll output" checkbox moved above the logs (#1417)

Capture d’écran 2026-03-04 à 14 01 43

# Issues this PR resolves
- Closes #2106
- Closes #1417

# A checklist for hand testing
- [x] Make submission and check the interface
- [x] Add many fact sheet elements and check the interface

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] CircleCi tests are passing
- [x] Ready to merge

PR #2223, opened 2026-02-28
Compute worker - Improve status update and logs

Didayolo:
# Description

1. Make `_update_status` robust to failures
2. Log traceback instead of uninformative messages
3. Change status to `Failed` for any exception, avoiding getting stuck to `Running`
4. Add a "best effort" logging with `push_logs` method
5. Add a container removal in case of failure, avoid orphan containers
6. Some minor fixes
7. Improve `show_progress` clarity and avoid overlogging
8. Change in `tasks.py`: queue the compute worker task only after commit to avoid fetching too soon and raising a 404 error

Generally, this change should fix the "stuck in running" bug, and allow more logs to reach the platform's front-end. Possibly fix also the orphan containers issue.

# Issues this PR resolves
- #1205, potentially points 1, 2, partially 3, partially 4 and 5

# A checklist for hand testing
- [x] Check successful and failing submissions
- [x] Check logs inside container and on the platform

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] CircleCi tests are passing
- [x] Ready to merge

PR #2254, opened 2026-03-10
Merge Develop into Master (v1.25)

ObadaS:
# Upgrade instructions

It is required to re-build the containers:

```sh
docker compose build --no-cache
docker compose up -d
```

# Release notes

## New Features

- #2267
- #2219 
- #2055 (+ #2268)

## Improvements

- #2200 
- #2223 
- #2245
- #2256 
- #2238
- #2272
- #2270
- #2273
- #2280 
- #2281

## Bug Fix

- #2239
- #2221  
- #2228 
- #2257 
- #2287 
- #2276 

## Security

- #2264 
- #2260 
- #2292 

## Documentation

- #2252 
- #2278


Didayolo:
We had the following display problem when loading pages. I suspect the change in `packaging/container/Containerfile` from #2055 and think we should revert the PR merge.

https://github.com/user-attachments/assets/005dbc18-f831-4c67-a6e2-a2e096858f18

ObadaS:
> We had the following display problem when loading pages. I suspect the change in `packaging/container/Containerfile` from #2055 and think we should revert the PR merge.
> Enregistrement.de.l.ecran.2026-03-17.a.20.20.01.mov


I can't reproduce this problem on my local instance.

Could this problem be Mac specific ? I tried with DEBUG=False and DEBUG=True

Didayolo (approved):

PR #2256, opened 2026-03-11
Support for Multiple Domain names (instance behind reverse proxy)

ObadaS:
# Description
If the instance is deployed behind a reverse proxy, which then proxies the requests via HTTPS, there can be some errors related to CSRF which are fixed in this PR.

If the DOMAIN_NAME is not the same as the external domain name for the website, some things like emails will use an unknown domain name as the URL, this PR fixes this.

To fix these problems, use the new `EXTERNAL_DOMAIN_NAME` variable in the .env, setting it to the domain name that is available externally while keeping the `DOMAIN_NAME` locally accessible.

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/documentation/docs/Developers_and_Administrators/How-to-deploy-Codabench-on-your-server.md
+++ b/documentation/docs/Developers_and_Administrators/How-to-deploy-Codabench-on-your-server.md
@@ -440,3 +440,6 @@ Caddyfile :
     }
     reverse_proxy @min_bucket minio:{$MINIO_PORT}
 ```
+
+## Codabench Instance behind a reverse proxy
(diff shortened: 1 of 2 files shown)


Didayolo:
Is this related?

- #2213

ObadaS:
> Is this related?
> 
> * [Allow custom domain to be used with public S3 bucket #2213](https://github.com/codalab/codabench/pull/2213)

Not really, this is for the Caddy and Django containers when they are put behind a reverse proxy, if you want to use a different domain name internally and externally

Didayolo (approved):

PR #2267, opened 2026-03-16
Autogenerate SECRET_KEY or use given one

ObadaS:
Automatically generate Django's Secret Key if it does not exist inside the .env file, otherwise use the given secret key

Alternative to #2232 

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/.env_circleci
+++ b/.env_circleci
@@ -1,5 +1,3 @@
-SECRET_KEY=change-this-secret
-
 DB_HOST=db
 DB_NAME=postgres
 DB_USERNAME=postgres
--- a/.env_sample
+++ b/.env_sample
@@ -1,7 +1,5 @@
-SECRET_KEY=change-this-secret
-
 # For local setup and debug
-DEBUG=True
+DEBUG=False
 
 # Database
 DB_HOST=db
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,6 +66,7 @@ dependencies = [
     "django-filter==25.1",
     "django-cors-headers==4.9.0",
     "nh3==0.3.3",
+    "configobj==5.0.9",
 ]
 
 [tool.uv]
--- a/src/settings/base.py
+++ b/src/settings/base.py
@@ -137,14 +137,13 @@ with open(".env", "a+") as f:
     secret_key_count = 0
     f.seek(0)
     for x in f:
-        if "SECRET_KEY=" in x and x[0] == "S":
+        if x.strip().startswith("SECRET_KEY="):
             secret_key_count = 1
             SECRET_KEY = os.environ.get("SECRET_KEY", config['SECRET_KEY'])
             break
     if secret_key_count == 0:
         SECRET_KEY = get_random_secret_key()
-        f.write(f"SECRET_KEY='{SECRET_KEY}'")
-f.close()
+        f.write(f"\nSECRET_KEY={SECRET_KEY}\n")
 
 LOGIN_REDIRECT_URL = '/'
 LOGOUT_REDIRECT_URL = '/'
--- a/src/settings/base.py
+++ b/src/settings/base.py
@@ -143,7 +143,7 @@ with open(".env", "a+") as f:
             break
     if secret_key_count == 0:
         SECRET_KEY = get_random_secret_key()
-        f.write(f"\nSECRET_KEY={SECRET_KEY}\n")
+        f.write(f"\nSECRET_KEY='{SECRET_KEY}'\n")
 
 LOGIN_REDIRECT_URL = '/'
 LOGOUT_REDIRECT_URL = '/'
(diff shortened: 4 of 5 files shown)


Didayolo:
We need to update the documentation to indicate that admins should add a `SECRET_KEY` variable in `.env`

ObadaS:
The variable is added automatically in the .env file at launch if no secret key is defined.
But I can also update the documentation to document these changes

Didayolo:
> The variable is added automatically in the .env file at launch if no secret key is defined. But I can also update the documentation to document these changes

Ah yes I forgot about this. I thought it was going to be re-generated everytime instead of being written. Not sure if we want to document it then.

Didayolo:
@ObadaS 

Why do we use `config = configobj.ConfigObj('.env')` ?

I feel like it just adds a useless package. We are already manually parsing the `.env` file.

I think we could simply checks for the environment variable.

Maybe something like:

```python
env_secret = os.environ.get("SECRET_KEY")
if env_secret:
    SECRET_KEY = env_secret
else:
      SECRET_KEY = get_random_secret_key()
      with open(".env", "a") as f:
          f.write(f"\nSECRET_KEY={SECRET_KEY}\n")
```

---

Also, a nice call from #2232 is that it was updating `.env` at Docker entrypoint, not in `base/settings.py`, so it may be safer if other services use the key.

ObadaS:
> @ObadaS
> 
> Why do we use `config = configobj.ConfigObj('.env')` ?
> 
> I feel like it just adds a useless package. We are already manually parsing the `.env` file.
> 
> I think we could simply checks for the environment variable.

While we do read the lines in the files, it stores the whole line into the variable, so if we find the SECRET_KEY we would need to remove SECRET_KEY=' and the last ' before assigning it in Django


Also, I saw from your last commit that you removed `f.close()`, was that on purpose ?

Didayolo:
> > @ObadaS
> > Why do we use `config = configobj.ConfigObj('.env')` ?
> > I feel like it just adds a useless package. We are already manually parsing the `.env` file.
> > I think we could simply checks for the environment variable.
> 
> While we do read the lines in the files, it stores the whole line into the variable, so if we find the SECRET_KEY we would need to remove SECRET_KEY=' and the last ' before assigning it in Django
> 
> Also, I saw from your last commit that you removed `f.close()`, was that on purpose ?

I just edited my comment for more details.

About `f.close()` yes it is on purpose, you don't need it when you use `with open(".env", "a+") as f:`.

Didayolo:
By the way I tested it and it seems to work fine, I am just nitpicking on details

ObadaS:
> @ObadaS
> 
> Why do we use `config = configobj.ConfigObj('.env')` ?
> 
> I feel like it just adds a useless package. We are already manually parsing the `.env` file.
> 
> I think we could simply checks for the environment variable.
> 
> Maybe something like:
> 
> ```python
> env_secret = os.environ.get("SECRET_KEY")
> if env_secret:
>     SECRET_KEY = env_secret
> else:
>       SECRET_KEY = get_random_secret_key()
>       with open(".env", "a") as f:
>           f.write(f"\nSECRET_KEY={SECRET_KEY}\n")
> ```
> 
> Also, a nice call from #2232 is that it was updating `.env` at Docker entrypoint, not in `base/settings.py`, so it may be safer if other services use the key.


This code would not work since the base.py is loaded multiple times during launch (and everytime we do so Django commands through manage.py like shell_plus or migrations)

With your code, the secret key will keep getting written in the .env file until you restart the Django containers and make the secret key variable an environnement variable. So you end up with multiple secret keys in the .env 


As for #2232  putting the generation inside a docker entry point, it means that in the future, when we make Codabench binaries and don't rely on docker (or podman), the secret key generation won't work.

Didayolo:
@ObadaS OK, thank you for the explanation. On my side I think it is ready to be merged.

Didayolo (approved):

IdirLISN (commented):
seems fine to me. We will discuss more about it in team meeting

IdirLISN (approved):

ObadaS:
@Didayolo I add back the '' that you removed earlier. Their purpose was to fix a rare warning that manifested itself when launching the instance with docker depending on the composition of the secret key generated by Django.
image

Didayolo:
> @Didayolo I add back the '' that you removed earlier. Their purpose was to fix a rare warning that manifested itself when launching the instance with docker depending on the composition of the secret key generated by Django. image

OK, sorry for that

Didayolo:
> seems fine to me. We will discuss more about it in team meeting

So do you want to wait for the meeting before merging this PR?

IdirLISN:
> > seems fine to me. We will discuss more about it in team meeting
> 
> So do you want to wait for the meeting before merging this PR?

no, we can merge.

PR #2285, opened 2026-03-21
Compute Worker - Fix submission files duplication

ihsaan-ullah:
# Description
This PR has the following updates:
### `.env_sample`
- updated for minor changes to `WORKER_BUNDLE_URL_REWRITE`

### `src/apps/competitions/tasks.py`
- updated to send submission during both ingestion and scoring, and some other minor updates for clarity

### `computer_worker.py`
- separate class `SubmissionStatus` for submission statuses and related updates
- separate class `ProgramKind` for program kind and related updates
- introduced a new class `Settings` that gathers all env variables. This class is now used all over the code for accessing settings. NOTE: This can be further improved to convert strings to booleans so that we don't use `== "true"` or `== "false". 
- organized imports
- renamed `program` to `scoring_program`
- update code `start` function to clarify which functions will run during ingestion and during scoring, and separated submission and scoring program clearly
- made submission available ragardless of ingestion or scoring in `app/ingested_program`
- rewrote `_run_program_directory` function to simplify code and logs
- moved some code from `_run_program_directory` to a new function `_create_container` for reusability and clarity
- minor changes to `watch_detailed_results` function to avoid looping forever
- replaced `container_id` (undeclared) by `container.get("Id") in `_run_container_engine_cmd`
- [TO DISCUSS] commented some code (that was introduced in #2030 ) at the end of the `start` function that was failing submission even if everything went well. Need to discuss this and update the code

# Issues this PR resolves
- #1874

# A checklist for hand testing
- [ ] Upload this modified mini autoML bundle to check in ingestion logs and scoring logs that submission files are accessible [Modified_Bundle.zip](https://github.com/user-attachments/files/26165486/Modified_Bundle.zip)
- [ ] use this submission to test the bundle [submittion.zip](https://github.com/user-attachments/files/26165488/submittion.zip)
- [ ] Carefully check the code to make sure there are no bugs

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [ ] Code review by reviewer
- [ ] Hand tested by reviewer
- [ ] CircleCi tests are passing
- [ ] Ready to merge

PR #2293, opened 2026-03-26
ENV - added / at the end of the url

ihsaan-ullah:
# Description
Added `/` to the end of the url in env sample. Without this change the url is incorrect when used in the compute worker

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/.env_sample
+++ b/.env_sample
@@ -71,7 +71,7 @@ AWS_STORAGE_PRIVATE_BUCKET_NAME=private
 AWS_S3_ENDPOINT_URL=http://minio:9000/
 AWS_QUERYSTRING_AUTH=False
 # Optional URL rewriting in compute worker, format: FROM | TO
-#WORKER_BUNDLE_URL_REWRITE=http://localhost:9000|http://minio:9000
+#WORKER_BUNDLE_URL_REWRITE=http://localhost:9000/|http://minio:9000/
 
 
 # -----------------------------------------------------------------------------


Didayolo:
Is this related to the bug I reported in #2285? Or a bug you have locally?

I never had problem with this.

Didayolo (approved):

ihsaan-ullah:
> Is this related to the bug I reported in #2285? Or a bug you have locally?
> 
> I never had problem with this.

This is not a reported bug. I think you added a slash `/` in your version of `.env`. I added it in this PR so that users know that a slash at the end is required in order to have a working url e.g. `http://minio:9000/private` is valid but `http://minio:9000private` is not valid

Didayolo:
> > Is this related to the bug I reported in #2285? Or a bug you have locally?
> > I never had problem with this.
> 
> This is not a reported bug. I think you added a slash `/` in your version of `.env`. I added it in this PR so that users know that a slash at the end is required in order to have a working url e.g. `http://minio:9000/private` is valid but `http://minio:9000private` is not valid

Actually my config is:

```
WORKER_BUNDLE_URL_REWRITE=http://docker.for.mac.localhost:9000|http://minio:9000
```

Anyway I think it is more robust with the new `.env_sample`

PR #2310, opened 2026-04-01
Cancel Submission - use submission queue instead of competition queue

### LINKED ISSUE #872: Cancel running submissions
When we cancel a running submission, does it really get killed inside the worker? Or is it just a front-end move?

In CodaLab Competitions, we had this known that we can't kill running submissions from the interface. The only way is to go inside the compute workers and kill the submissions by hand using `docker stop` or `docker kill`.

Do we have the same limitation on Codabench? I guess the problem is not that simple because the worker stops listening to the queue once a job has started, and the platform has no possible actions on the worker.

ihsaan-ullah:
# Description
Originally we were using competition queue to cancel submissions. But we know that submission can have different queue than the competition. This PR changes that and uses submission queue.

# Issues this PR resolves
- #872

- [ ] We should test this on custom queue
- [ ] Run submission on `queue_1` then change the competition queue to `queue_2` and cancel the submission 

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/src/apps/competitions/models.py
+++ b/src/apps/competitions/models.py
@@ -674,8 +674,10 @@ class Submission(models.Model):
                     sub.cancel(status=status)
             celery_app = app
             # If a custom queue is set, we need to fetch the appropriate celery app
-            if self.phase.competition.queue:
-                celery_app = app_for_vhost(str(self.phase.competition.queue.vhost))
+            # NOTE: We fetch the queue from submission and not from competition to be sure that we are using the correct queue
+            # Originally we were using queue from the competition (self.phase.competition.queue)
+            if self.queue:
+                celery_app = app_for_vhost(str(self.queue.vhost))
             # We need to convert the UUID given by celery into a byte like object otherwise it won't work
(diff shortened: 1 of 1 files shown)


Didayolo:
Are we sure the cancelation work after this change and #872 should be closed?

Sometimes we never changed the queue (typically default queue) and from what I remember the run continues until the end even when we try to cancel it. We need to test this.

ihsaan-ullah:
I checked on my side with iris and the cancellation worked. 

Even if cancellation does not work, this change makes everything safe because let say you submit a submission and the competition queue is `queue_A`. while the submission is running you change the comepittion queue to `queue_B`. Now if you want to cancel, cancel request will be sent to `queue_B` which is wrong. The changes i made will make sure that if submission is using `queue_A` then cancel request will be sent to `queue_A`

Didayolo (approved):

PR #2313, opened 2026-04-02
Quota Cleanup - Updates to remove unused starting kits and bundles and exclude starting kit and public data from unused datasets

### LINKED ISSUE #2277: Does "Delete unused Datasets and Programs" delete public datasets?
We need to check if the following option "Delete unused Datasets and Programs":

Image

Removes public datasets that are linked to no tasks but listed in public datasets page: https://www.codabench.org/datasets/

If it is the case, we should fix this behavior.

@ihsaan-ullah

ihsaan-ullah:
# Description
This PR updates the quota cleanup functionality to:
- exclude starting kits and public data from `unsued_datasets`
- enable users to remove `unused_starting_kits` and `unused_competition_bundles`
- tests are updated for the new additions and updates

# Screenshot:
Screenshot 2026-04-02 at 8 26 02 AM

# Issues this PR resolves
- #2277

- [x] test carefully to check if quota cleanup section stats are correct and that deletion works as expected.

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge

### DIFF
--- a/src/apps/api/tests/test_cleanup.py
+++ b/src/apps/api/tests/test_cleanup.py
@@ -56,8 +56,19 @@ class CleanUpTests(APITestCase):
             DataFactory(created_by=user, type=Data.INGESTION_PROGRAM),
             DataFactory(created_by=user, type=Data.SCORING_PROGRAM),
             DataFactory(created_by=user, type=Data.INPUT_DATA),
-            DataFactory(created_by=user, type=Data.REFERENCE_DATA),
-            DataFactory(created_by=user, type=Data.PUBLIC_DATA)
+            DataFactory(created_by=user, type=Data.REFERENCE_DATA)
+        ]
+
+        # Create unused starting kits
+        self.unused_starting_kits = [
+            DataFactory(created_by=user, type=Data.STARTING_KIT),
+            DataFactory(created_by=user, type=Data.STARTING_KIT)
+        ]
+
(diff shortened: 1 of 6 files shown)


Didayolo:
When rebasing on `develop` and solving conflict, instead of simply `git add` the `bundle_management.tag` file, I used a shady VSCode feature. It pushed files in a strange way. I'm manually fixing this.

Didayolo:
Note about #2277 issue:

The problem is way worse than what I thought. The option "Delete unused datasets and programs" actually **remove all your starting kits and public datasets, even if they are marked as IN USE**. This is a very destructive bug.

@ihsaan-ullah Is your PR solving this as a by-product?

ihsaan-ullah:
You are right, the `In Use` is different for datasets/programs and public data and starting kits.

Now when these are separated, this destructive issue is also resolved with this condition:
```python
Q(competition__isnull=True) &
Q(phase_starting_kit__isnull=True)
```
which means to only count those `Data` in `unused starting kits and bundles` whose competition is null and are not attached to any phase. 

NOTE: now we do not care about unused public data. If someone wants to delete them they will have to go to the datasets tab and delete them manually.

Didayolo (approved):

PR #2377, opened 2026-05-15
Compute Worker - Docker image update check added

### LINKED ISSUE #2320: Computer worker upgrade feature
We need a feature in compute worker to communicate with codabench and check for any critical or non critical upgrades available
- If a critical upgrade is found, compute worker will fail with a nice message
- If the upgrade is not critical then computer worker prints a warning
- if no upgrade just execute normally

ihsaan-ullah:
# Description
- This PR adds additional code to compute worker to compare local and remote compute worker images
- It also has small update to circle-ci config to run flake-8 on compute worker directory

# Logs
Compute worker container will show logs like below when this new code runs
Screenshot 2026-05-16 at 1 53 45 AM

# Issues this PR resolves
- #2320

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [ ] Hand tested by reviewer
- [ ] Ready to merge

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,7 +66,9 @@ jobs:
 
       - run:
           name: "Lint: Check code style with flake8"
-          command: docker compose exec django flake8 src/
+          command: |
+            docker compose exec django flake8 src/
+            docker compose exec django flake8 compute_worker/
 
       - run:
           name: "Tests: Run unit/integration tests (excluding e2e)"
--- a/compute_worker/compute_worker.py
+++ b/compute_worker/compute_worker.py
@@ -317,7 +317,7 @@ def check_docker_image_update():
         tag=Settings.DOCKER_IMAGE_TAG,
         docker_base_url=Settings.CONTAINER_SOCKET
     )
-    result = checker.compare_local_vs_remote_images()
+    result = checker.compare_local_vs_remote_images(container_id=socket.gethostname())
     status = result["status"]
 
     log_level = logging.INFO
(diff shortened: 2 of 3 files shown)


Didayolo:
So this adds warning in the worker logs when the version is not up-to-date?

- Do we have anything visible on the platform's UI? Like, the logs appear in the submission logs?
- Can we differentiate mandatory updates from optional updates? (that is maybe not that useful)

ihsaan-ullah:
> So this adds warning in the worker logs when the version is not up-to-date?

Yes 

> * Do we have anything visible on the platform's UI? Like, the logs appear in the submission logs?

No, this will only be visible in compute worker logs

> * Can we differentiate mandatory updates from optional updates? (that is maybe not that useful)

For now this will only check if local is behind reemote or not. Later on if needed we can somehow differentiate between critical and non-critical updates. This first version do not interact with anything except the docker image on remote(docke hub)

Didayolo:
I think the warning should be visible on platform side, typically in _ingestion_error_logs_ and _scoring_error_logs_.

For the "optional or mandatory" update I think it is not that important.

Didayolo:
Update:

- In case of problem we write logs in "scoring error logs" and "ingestion error logs"
- We check by current container ID instead of just the docker image

The reason for this second point is that, otherwise, if you `git pull` the latest image but do not restart the container, it will be marked as up-to-date.

The problem with this approach is that in many case the container won't be found by `inspect_container` just by container ID (typically because of the format "projectname_containername"), so it will throw an error and go back to only checking the image.

PR #2395, opened 2026-05-28
Worker status/monitoring for competition organizers (from rabbit)

IdirLISN:
/!\ New PR with modified feature. /!\

The feature is now aware of workers health directly from django to rabbit instead of going through the site worker which slows the app massively. Using rabbit enable the feature to be much more responsive and display the worker state faster. the implementation is easier also. 

check the feature at: https://dev.codabench.org/
------------------------------- OLD PR (edited) --------------------------------------
# @ mention of reviewers
@ObadaS 
@Didayolo 
@ihsaan-ullah 

# Description

This feature enables admin users, competition organisers and his/her collaborators to monitor compute worker activity.
Public compute workers displayed by default and if the competition own a private queue, it displays the private compute workers linked to the queue of the competition. 

UI inside of the header of the competition page.

Feature contains: 
Feature available only for competition organizers and collaborators.
When disabled or hided behind the button, the websocket connection is disabled for performance.

image

image

# Issues this PR resolves
Enables organizers to monitor compute workers activity to push them to use their own workers in case of high activity on codabench public compute workers.

# A checklist for hand testing
- [x] Try the panel as a organizer, you should be able to see it by pressing the button.
- [x] Submit and watch if the panel shows the activity of the compute worker
- [x] Verify websocket state when the panel is not displayed, the connection should be deactivated. 
       (docker compose logs -f django should display websocket open and close with the panel) 

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] CircleCi tests are passing
- [x] Ready to merge

PR #2408, opened 2026-06-15
fix(submissions): prevent silent submission losses

AybH26:
## Title
fix(submissions): prevent silent submission losses

## Summary
Fixes the silent submission-loss incident first reported in 2025 (~30 %
of submissions stuck in `Submitting`/`Submitted`/`Scoring` with no
traceable error).
---

## Root causes

The incident wasn't one bug — it was a chain of five compounding issues
across the Django site worker, Celery beat, and the compute worker.
Each could silently drop a submission on its own; combined they turned
idle traffic into a ~30 % loss rate.

| ID   | Layer            | Bug                                                                                                                                   |
| ---- | ---------------- | ------------------------------------------------------------------------------------------------------------------------------------- |
| F1   | site_worker      | `watchmedo` wrapped Celery prefork → SIGTERM not forwarded → channel torn down without acking the in-flight `run_submission`          |
| F2   | beat             | `CELERY_BEAT_SCHEDULE` published `refresh_compute_worker_health` (does not exist) → `KeyError` killed the consumer channel            |
| F3   | site_worker      | Memory limit 256 MB was OOM-killed under prefork(2) + beat → orphan `Submitting` rows, `celery_task_id=None`                          |
| M1.A | Django watchdog  | No recovery path for submissions stuck in `Scoring`/`Running` after a worker crash — they stayed stuck forever                        |
| M1.B | compute_worker   | (a) `urllib3.Retry` excludes PATCH by default — the final `PATCH status=Finished` was never retried. (b) `_update_status` swallowed every exception, so `task_acks_late=True` acked tasks whose terminal update had failed. |

---

### F1 — `docker-compose.yml`
- Drop the `watchmedo` wrapper around the Celery prefork pool on `site_worker`.
- Subscribe to both `site-worker` and the default `celery` queue so beat-published tasks land where the worker consumes.

### F2 — `src/settings/base.py`
- Comment out the `refresh_compute_worker_health` beat entry (task does not exist anywhere in the codebase).

### F3 — `docker-compose.yml`
- Raise `site_worker` memory limit from `256M` to `1G`.

### M1.A — `src/apps/competitions/tasks.py` + `src/settings/base.py`
- New `reaper_stuck_scoring` Celery task on `site-worker` queue, soft limit 5 min.
- Re-dispatches any submission stuck in `Scoring`/`Running` past `started_when + execution_time_limit + threshold_minutes`.
- Scheduled every 5 min in `CELERY_BEAT_SCHEDULE` with `threshold_minutes=30`.
- Annotates `Submission.status_details` with a reaper marker for `/admin` and API visibility.

### M1.B — `compute_worker/compute_worker.py`
- `requests` session adapter: `Retry(total=5, backoff_factor=2, status_forcelist=(500,502,503,504), allowed_methods=frozenset((..., PATCH)))`.
- `_update_status` now re-raises when `status == FINISHED` so Celery (with `task_acks_late=True`) requeues the job.
  Intermediate states (`Running`, `Scoring`) stay best-effort to avoid killing a scoring run for a transient glitch.
  `Failed` already has a `raise` in every caller, so the requeue path is covered.

---

### DIFF
--- a/compute_worker/compute_worker.py
+++ b/compute_worker/compute_worker.py
@@ -499,12 +499,19 @@ class Run:
         websocket_scheme = "ws" if submission_api_url_parsed.scheme == "http" else "wss"
         self.websocket_url = f"{websocket_scheme}://{websocket_host}/submission_input/{self.user_pk}/{self.submission_id}/{self.secret}/"
 
-        # Nice requests adapter with generous retries/etc.
+        # M1.B: urllib3.Retry skips PATCH by default. The submission status
+        # callback uses PATCH, so without explicitly allowing it any 5xx /
+        # transient timeout on the final PATCH status=Finished was lost,
+        # leaving the submission stuck in Scoring forever.
         self.requests_session = requests.Session()
         adapter = requests.adapters.HTTPAdapter(
             max_retries=Retry(
-                total=3,
-                backoff_factor=1,
+                total=5,
+                backoff_factor=2,
+                status_forcelist=(500, 502, 503, 504),
+                allowed_methods=frozenset(
+                    ("GET", "POST", "PUT", "DELETE", "HEAD", "OPTIONS", "PATCH")
+                ),
             )
         )
         self.requests_session.mount("http://", adapter)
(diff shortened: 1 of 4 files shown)


Didayolo:
Hi @AybH26,

Thank you for your contribution.

What was the protocol used to reproduce the problem, and to confirm that this change solves it?

To my knowledge, this problem was solved by #2223.

IdirLISN:
"CELERY_BEAT_SCHEDULE published refresh_compute_worker_health (does not exist) → KeyError killed the consumer channel" 

This is correct, it's an artefact from the last compute worker monitoring deployment and this is fixed in the next PR for worker monitoring (#2395).

Didayolo:
Hello @AybH26,

I made some comments but I'll give general directions here:

- Please clean up the comments to keep only the relevant comments (avoid references to your naming of the issue, or overkill explanations)

- Please separate the pull request into sub-PR. Each sub-PR should tackle one clear point (e.g. increase the site_worker memory, add a new task, etc.)

As always, you can contact us directly if you have any doubt.

PR #2416, opened 2026-06-16
fix: add retry with backoff to _update_submission and stop swallowing…

### LINKED ISSUE #2415: _update_submission() has no retry and _update_status() silently swallows exceptions, causing permanent submission loss
## Bug

`_update_submission()` in `compute_worker/compute_worker.py` performs a single HTTP PATCH to report submission status back to Django. On any non-200 response, it raises `SubmissionException` immediately with no retry:

```python
resp = self.requests_session.patch(url, data=data, timeout=150)
if resp.status_code == 200:
    logger.info("Submission updated successfully!")
else:
    raise SubmissionException("Failure updating submission data.")
```

The caller `_update_status()` then catches **all** exceptions and silently discards them:

```python
try:
    self._update_submission(data)
except Exception as e:
    # Always catch exception and never raise error
    logger.exception(f"Failed to update submission status to {status}: {e}")
```

This means a single transient network error (502, timeout, connection reset) permanently loses the submission result. The compute worker moves on to the next job as if nothing happened, while Django never receives the `status=Finished` PATCH. The submission stays stuck in `Scoring` or `Running` forever.

## How it happens

1. Compute worker finishes scoring and sends `PATCH /api/submissions/{id}/ {status: Finished}`
2. The network returns a transient error (502, timeout, connection reset)
3. `_update_submission()` raises `SubmissionException("Failure updating submission data.")`
4. `_update_status()` catches the exception, logs it, and **does not re-raise**
5. The compute worker proceeds to the next submission
6. Django never receives the `Finished` status → the submission stays stuck in `Scoring` indefinitely
7. The participant sees a perpetually "processing" submission with no error message and no way to know it failed

dconstancy:
**Title:** fix: add retry with backoff to `_update_submission()` and stop swallowing terminal exceptions

---

# Description

`_update_submission()` performs a single PATCH request with no retry. On any transient network error (502, timeout, connection reset), the submission result is permanently lost. The caller `_update_status()` silently swallows all exceptions, so the compute worker moves on while Django never receives the final status. The submission stays stuck in `Scoring` or `Running` forever with no user-facing error.

This PR adds exponential-backoff retry to `_update_submission()` and re-raises exceptions for terminal statuses so submissions no longer hang silently.

# Issues this PR resolves

Fixes #2415

# Changes

**`compute_worker/compute_worker.py`:**

- `_update_submission()`: add retry loop with exponential backoff (5 attempts, 2^n seconds + jitter). Catches both HTTP error responses and network-level `RequestException`. Only raises after all retries are exhausted.
- `_update_status()`: re-raise exceptions for terminal statuses (`Finished`, `Failed`) so Celery marks the task as failed and the submission transitions to `Failed` in the UI instead of hanging silently. Intermediate statuses (`Preparing`, `Running`, `Scoring`) remain silently caught — losing an intermediate status update is not critical.
- HTTP session adapter: increase `total` retries from 3 to 5, add `status_forcelist=[502, 503, 504]` and `allowed_methods=["PATCH", "GET", "PUT", "POST"]` to automatically retry on server errors at the transport level.

- [ ] Trigger a submission and observe successful status updates in compute worker logs
- [ ] Simulate a 502 response (e.g. restart Django mid-submission) and confirm retry succeeds
- [ ] Confirm submissions reach `Finished` or `Failed` state and no longer hang in `Scoring`/`Running`

# Backward compatibility

- No API changes, no migration needed
- Retry is transparent to Django — same PATCH payload, just sent multiple times if needed
- PATCH to `/api/submissions/{id}/` is idempotent (sets status to a fixed value), so retries are safe
- Intermediate status exceptions remain silently caught (no behavior change for non-terminal updates)
- Only terminal status failures (`Finished`, `Failed`) now propagate — this is strictly better than the previous silent loss

# Checklist

- [x] Code review by me
- [x] Hand tested by me
- [ ] Code review by reviewer
- [ ] Hand tested by reviewer
- [ ] Ready to merge

### DIFF
--- a/compute_worker/compute_worker.py
+++ b/compute_worker/compute_worker.py
@@ -8,6 +8,7 @@ import shutil
 import signal
 import socket
 import tempfile
+import random
 import time
 import uuid
 import requests
@@ -503,8 +504,10 @@ class Run:
         self.requests_session = requests.Session()
         adapter = requests.adapters.HTTPAdapter(
             max_retries=Retry(
-                total=3,
-                backoff_factor=1,
+                total=5,
+                backoff_factor=2,
+                status_forcelist=[502, 503, 504],
+                allowed_methods=["PATCH", "GET", "PUT", "POST"],
             )
         )
         self.requests_session.mount("http://", adapter)
@@ -614,20 +617,36 @@ class Run:
             ]
         return [run_args[name] for name in DETAILED_OUTPUT_NAMES]
 
-    def _update_submission(self, data):
+    def _update_submission(self, data, max_retries=5, backoff_base=2):
         url = f"{self.submissions_api_url}/submissions/{self.submission_id}/"
         data["secret"] = self.secret
 
-        logger.info(f"Updating submission @ {url} with data = {data}")
+        for attempt in range(1, max_retries + 1):
+            logger.info(f"Updating submission @ {url} (attempt {attempt}/{max_retries}) with data = {data}")
+            try:
+                resp = self.requests_session.patch(url, data=data, timeout=150)
+            except requests.exceptions.RequestException as exc:
+                logger.warning(f"Submission patch request failed (attempt {attempt}/{max_retries}): {exc}")
+                if attempt == max_retries:
+                    raise SubmissionException(f"Failure updating submission data after {max_retries} attempts.")
+                sleep_s = backoff_base ** attempt + random.uniform(0, 1)
+                logger.info(f"Retrying in {sleep_s:.1f}s...")
+                time.sleep(sleep_s)
+                continue
 
-        resp = self.requests_session.patch(url, data=data, timeout=150)
-        if resp.status_code == 200:
-            logger.info("Submission updated successfully!")
-        else:
-            logger.error(
-                f"Submission patch failed with status = {resp.status_code}, and response = \n{resp.content}"
-            )
-            raise SubmissionException("Failure updating submission data.")
+            if resp.status_code == 200:
+                logger.info("Submission updated successfully!")
+                return
+            else:
+                logger.warning(
+                    f"Submission patch failed (attempt {attempt}/{max_retries}) "
+                    f"with status = {resp.status_code}, and response = \n{resp.content}"
+                )
+                if attempt == max_retries:
+                    raise SubmissionException(f"Failure updating submission data after {max_retries} attempts.")
+                sleep_s = backoff_base ** attempt + random.uniform(0, 1)
+                logger.info(f"Retrying in {sleep_s:.1f}s...")
+                time.sleep(sleep_s)
 
     def _update_status(self, status, extra_information=None):
         # Update submission status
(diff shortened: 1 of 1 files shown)


IdirLISN:
Hello, i reviewed the code, i think the changes are good but some details seems strange like:

```
        url = f"{self.submissions_api_url}/submissions/{self.submission_id}/"
        data["secret"] = self.secret

        logger.info(f"Updating submission @ {url} with data = {data}")
``` 
This seems to print some secret in the logs.

Secondly it seems you use a custom retry but urllib retry is alreay imported, also you add inside of the allowed methods, the method post:               ```  allowed_methods=["PATCH", "GET", "PUT", "POST"] ```
This shouldn't be there and the request should be hard implemented. This reduce the risk of post abuse.

dconstancy:
Hello, thanks for the feedback! I've updated the PR accordingly:

- The secret was already present in the logged data — removed it from the log line in `_update_submission`.
- Dropped the custom retry loop — no point in reinventing it when `urllib3.Retry` already does the job at the session level.
- Removed `POST` from `allowed_methods` since we don't retry POST calls.

compute_worker/compute_worker.py
```diff
@@ -618,14 +620,15 @@ def _update_submission(self, data):
...
+        logger.info(f"Updating submission @ {url}")
 
         resp = self.requests_session.patch(url, data=data, timeout=150)
         if resp.status_code == 200:
             logger.info("Submission updated successfully!")
         else:
             logger.error(
-                f"Submission patch failed with status = {resp.status_code}, and response = \n{resp.content}"
+                f"Submission patch failed with status = {resp.status_code}, "
+                f"and response = \n{resp.content}"
```
wlln: Please don't lint old code

compute_worker/compute_worker.py
```diff
@@ -639,8 +642,9 @@ def _update_status(self, status, extra_information=None):
         try:
             self._update_submission(data)
         except Exception as e:
-            # Always catch exception and never raise error
```
wlln, line 642: Please update the comment instead of removing it

compute_worker/compute_worker.py
```diff
@@ -503,8 +503,10 @@ def __init__(self, run_args):
         self.requests_session = requests.Session()
         adapter = requests.adapters.HTTPAdapter(
             max_retries=Retry(
-                total=3,
```
wlln, line 506: Why were `total` and `backoff_factor` modified ?

wlln (changes requested):

compute_worker/compute_worker.py
dconstancy: Unintentional reformatting, reverted.

compute_worker/compute_worker.py
dconstancy: Updated to reflect the new behavior: exceptions are now only re-raised for terminal statuses (Finished, Failed) so Celery marks the task as failed. Non-terminal status failures are still swallowed.

Add the following comment : **# Re-raise only for terminal statuses so Celery marks the task as failed.**

compute_worker/compute_worker.py
dconstancy: `total` and `backoff_factor` were bumped for an earlier version that had an explicit retry loop — reverted since that loop was removed.

PR #2460, opened 2026-07-03
Option to stop compute worker from sending logs and/or prediction files to Codabench

ObadaS:
- This PR adds the ability to stop the compute worker from sending the competition container logs and/or prediction files to Codabench. Instead, the logs will be saved locally on disk.
- Rename the folders that are created to be more descriptive and less random.
- Updated the documentation to show the new options in the compute worker .env

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge


Didayolo:
Hi @ObadaS,

Some points looks problematic to me:

### 1. Readability issue in local log file

In `push_logs`, `data` is `bytes`, so `f.write(str(data))` writes `b'...\n...'` on a single line instead of the actual text.

This is not critical but the logs produced will be annoying, with no line breaks.

Possible solution: open the file in `"wb"` mode and write the bytes directly (ideally with a header per program/stream).

### 2. Logs are deleted right after being written

With only `COMPUTE_WORKER_DISABLE_LOG_UPLOAD=True`, `push_logs()` writes the file inside `root_dir`, then `clean_up()` deletes `root_dir`. Maybe we could write them outside `root_dir` (e.g. `/codabench/logs/`) or force `NO_CLEANUP` like for predictions.

### 3. Local prediction transfer fails silently with multiple workers

I think in the doc it is said that this work only with 1 worker in the queue. But that is not explicitly reported in case of problem.

With `COMPUTE_WORKER_DISABLE_PREDICTION_UPLOAD`, scoring must run on the same worker as prediction, but nothing enforces or checks this. If the folder is missing, `copytree` raises a raw `FileNotFoundError`. 

I think we need to check that the folder exists and raise a clear `SubmissionException`.


### 4. Predictions copied `scoring_program` folder

Predictions are copied into `scoring_program/submission/`. From what I get, the folder contains the submission data sent during scoring, so code and predictions end up mixed in both `/app/ingested_program` and `/app/input/res`, no? Not sure about this point. But maybe the predictions could be copied directly into `input/res`, as usual.

Didayolo (approved):

PR #2491, opened 2026-08-05
Increase default queue limit to 10

Didayolo:
# Description

Users can now have up to 10 queues by default. We do not have good reasons to set to the limit to 5.

# Upgrade instructions

Django migration

When deployed we should update the value for old users to 10 using the following code:

Open shell plus
```python
docker compose exec django ./manage.py shell_plus
```

Update old users limit to 10
```python
from profiles.models import User
User.objects.all().update(rabbitmq_queue_limit=10)
```

# Checklist
- [x] Code review by me 
- [x] Hand tested by me 
- [x] I'm proud of my work
- [x] Code review by reviewer
- [x] Hand tested by reviewer
- [x] Ready to merge


ihsaan-ullah:
I tested this PR, it works fine. 

When deployed we should update the value for old users to 10 using the following code:


Open shell plus
```python
docker compose exec django ./manage.py shell_plus
```

Update old users limit to 10
```python
from profiles.models import User
User.objects.all().update(rabbitmq_queue_limit=10)
```

ihsaan-ullah (approved):

